PR #19, opened 2020-01-31
Filter audio and video codecs in dash

krystalmejia24:


pkg/filters/dash.go
```diff
@@ -58,43 +59,85 @@ func (d *DASHFilter) FilterManifest(filters *parsers.MediaFilters) (string, erro
 		manifest.BaseURL = baseURLWithPath(path.Join(path.Dir(u.Path), manifest.BaseURL))
 	}
 
+	for _, filter := range d.getFilters(filters) {
+		filter(filters, manifest)
```
zsiec: Love this refactor. 👍

pkg/filters/dash.go
```diff
@@ -58,43 +59,85 @@ func (d *DASHFilter) FilterManifest(filters *parsers.MediaFilters) (string, erro
...
-		supportedCaptionTypes[captionType] = struct{}{}
+		supportedCaptionTypes[string(captionType)] = struct{}{}
 	}
 
+	filterContentType(captionContentType, supportedCaptionTypes, manifest)
+}
+
+func filterContentType(filter ContentType, supportedContentTypes map[string]struct{}, manifest *mpd.MPD) {
 	for _, period := range manifest.Periods {
 		for _, as := range period.AdaptationSets {
```
zsiec: Question: should we be removing empty adaptation sets?

pkg/filters/dash.go
```diff
@@ -58,43 +59,85 @@ func (d *DASHFilter) FilterManifest(filters *parsers.MediaFilters) (string, erro
...
+						}
+						continue
 					}
-				}
 
+					for supported := range supportedContentTypes {
+						if ValidCodecs(*r.Codecs, CodecFilterID(supported)) {
+							filteredReps = append(filteredReps, r)
+						}
+					}
```
zsiec: Maybe we make this a `switch` statement where we do different execution for different filter types?

```go
switch filter {
  case captionContentType:
    // some filter looking for exact match
  case audioContentType, videoContentType:
    // some filter running `strings.Contains`
}
```

pkg/filters/dash_test.go
```diff
@@ -80,6 +80,314 @@ func TestDASHFilter_FilterManifest_baseURL(t *testing.T) {
...
+			if g, e := manifest, tt.expectManifestContent; g != e {
+				t.Errorf("FilterManifest() wrong manifest returned\ngot %v\nexpected: %v\ndiff: %v", g, e,
+					cmp.Diff(g, e))
+			}
+		})
+	}
+}
+
+func TestDASHFilter_FilterManifest_audioCodecs(t *testing.T) {
+	manifestWithEAC3AndEAC3AudioCodec := `<?xml version="1.0" encoding="UTF-8"?>
```
zsiec: should this be named `manifestWithEAC3AndAC3AudioCodec`?

pkg/filters/dash.go
krystalmejia24: inspired by mediahub filters! 😜

pkg/filters/dash_test.go
krystalmejia24: great catch. Just fixed.

pkg/filters/dash.go
krystalmejia24: I tested with Dash ref. player and it did playback just fine. But I've removed them anyway. I cant think of a reason to have an Adaptation Set without any Representations.

I have also updated the index so the IDs reflect the order in which they are present.

krystalmejia24:
I've made the changes discussed here and in person. thanks @zsiec 


I have one q though. Is it safe to assume the ID's for an AdaptationSet is the index? I'm not sure it matters and I am a ok with us restructuring the ID's coming out of bakery that way. I mostly ask bc I'm afraid there is some dependency somewhere with that. I feel like I have seen examples of AdaptationSet ID's that don't look like they reference any indexing of any kind

pkg/filters/dash.go
```diff
@@ -139,6 +183,26 @@ func (d *DASHFilter) filterAdaptationSetType(filters *parsers.MediaFilters, mani
...
+func isCodecSupported(codec string, ct ContentType, supportedCodecs map[string]struct{}) bool {
+	//the key in supportedCodecs for captionContentType is equivalent to codec
+	//advertised in manifest. we can avoid iterating through each key
+	if ct == captionContentType {
+		if _, found := supportedCodecs[codec]; found {
+			return true
+		}
+
+		return false
+	}
```
zsiec:
```suggestion
	if ct == captionContentType {
		_, found := supportedCodecs[codec]
		return found
	}

krystalmejia24: good eye!

Issue #17, opened 2020-01-24
Add filter for omitting video bitrates in Dash manifest


jigneshdhruv:

zsiec:
The existing bandwidth filter we have implemented for HLS needs to be ported to DASH.

PR #20, opened 2020-02-03
Bitrate Filter for DASH Manifests

sam-eng:
refs #17

### DIFF
--- a/pkg/filters/dash.go
+++ b/pkg/filters/dash.go
@@ -2,7 +2,6 @@ package filters
 
 import (
 	"fmt"
-	"math"
 	"net/url"
 	"path"
 	"strconv"
@@ -232,16 +231,6 @@ func (d *DASHFilter) filterBandwidth(filters *parsers.MediaFilters, manifest *mp
 		for index, as := range period.AdaptationSets {
 			as.ID = strptr(strconv.Itoa(index))
 		}
-	}
-}
-
-func bandwidthInRange(minBandwidth int, maxBandwidth int) bool {
-	if (minBandwidth >= 0 && maxBandwidth <= math.MaxInt32) &&
-		(minBandwidth < maxBandwidth) &&
-		!(minBandwidth == 0 && maxBandwidth == math.MaxInt32) {
-		return true
-	} else {
-		return false
 	}
 }
 
--- a/pkg/filters/dash.go
+++ b/pkg/filters/dash.go
@@ -2,6 +2,7 @@ package filters
 
 import (
 	"fmt"
+	"math"
 	"net/url"
 	"path"
 	"strconv"
@@ -231,6 +232,16 @@ func (d *DASHFilter) filterBandwidth(filters *parsers.MediaFilters, manifest *mp
 		for index, as := range period.AdaptationSets {
 			as.ID = strptr(strconv.Itoa(index))
 		}
+	}
+}
+
+func bandwidthInRange(minBandwidth int, maxBandwidth int) bool {
+	if (minBandwidth >= 0 && maxBandwidth <= math.MaxInt32) &&
+		(minBandwidth < maxBandwidth) &&
+		!(minBandwidth == 0 && maxBandwidth == math.MaxInt32) {
+		return true
+	} else {
+		return false
 	}
 }
 
--- a/pkg/filters/dash.go
+++ b/pkg/filters/dash.go
@@ -84,6 +84,10 @@ func (d *DASHFilter) getFilters(filters *parsers.MediaFilters) []execFilter {
 		filterList = append(filterList, d.filterCaptionTypes)
 	}
 
+	if filters.DefinesBitrateFilter() {
+		filterList = append(filterList, d.filterBandwidth)
+	}
+
 	return filterList
 }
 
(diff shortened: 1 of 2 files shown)


pkg/filters/dash.go
```diff
@@ -66,6 +67,10 @@ func (d *DASHFilter) FilterManifest(filters *parsers.MediaFilters) (string, erro
 		d.filterCaptionTypes(filters, manifest)
 	}
 
+	if bandwidthInRange(filters.MinBitrate, filters.MaxBitrate) {
```
zsiec: Should we rename this function to something like `isValidBitrateRange` to make it clearer that we're checking to make sure the requester has defined valid filters for us to run?

pkg/filters/dash.go
```diff
@@ -139,6 +144,47 @@ func (d *DASHFilter) filterAdaptationSetType(filters *parsers.MediaFilters, mani
...
 }
 
+func (d *DASHFilter) filterBandwidth(filters *parsers.MediaFilters, manifest *mpd.MPD) {
+	for _, period := range manifest.Periods {
+		var filteredAdaptationSets []*mpd.AdaptationSet
+
+		for _, as := range period.AdaptationSets {
+			if as.ContentType == nil {
+				continue
+			}
```
zsiec: I don't think we're using `ContentType` in this function, I think we can safely remove this.

pkg/filters/dash.go
```diff
@@ -139,6 +144,47 @@ func (d *DASHFilter) filterAdaptationSetType(filters *parsers.MediaFilters, mani
...
+
+			var filteredReps []*mpd.Representation
+
+			for _, r := range as.Representations {
+				if int(*(r.Bandwidth)) <= filters.MaxBitrate && int(*(r.Bandwidth)) >= filters.MinBitrate {
```
zsiec: Maybe we should change the `MaxBitrate` and `MinBitrate` values to be `int64` and avoid a type conversion here?

Also, one note for safety, because the `Representation` has a `Bandwidth` property that's a int64 pointer, if we simply dereference the pointer without asserting it's not `nil`, it can trigger a panic and halt execution unexpectedly. Here's an example of this in action: https://play.golang.org/p/FClazLnQrxS.

To fix this, we can do a simple check for each representation like:

```go
if r.Bandwith == nil {
  continue // we can't run any filtering logic on a representation that doesn't contain a bandwidth value
}
// continue with filter logic and dereference with confidence it won't trigger a panic
```

pkg/filters/dash_test.go
```diff
@@ -300,3 +301,145 @@ func TestDASHFilter_FilterManifest_filterStreams(t *testing.T) {
...
+      <Representation bandwidth="4096" codecs="avc" id="1"></Representation>
+    </AdaptationSet>
+    <AdaptationSet id="1" lang="en" contentType="audio">
+      <Representation bandwidth="256" codecs="ac-3" id="0"></Representation>
+    </AdaptationSet>
+  </Period>
+</MPD>
+`
+
+	filteredManifestHittingUpperBound := `<?xml version="1.0" encoding="UTF-8"?>
```
zsiec: Suggesting naming updates for test clarity, just some ideas if they're useful!

Perhaps this one could be `manifestFiltering256And2048Representations`

pkg/filters/dash_test.go
```diff
@@ -300,3 +301,145 @@ func TestDASHFilter_FilterManifest_filterStreams(t *testing.T) {
...
+  <BaseURL>http://existing.base/url/</BaseURL>
+  <Period>
+    <AdaptationSet id="0" lang="en" contentType="video">
+      <Representation bandwidth="4096" codecs="avc" id="1"></Representation>
+    </AdaptationSet>
+  </Period>
+</MPD>
+`
+
+	filteredManifestWithBothAS := `<?xml version="1.0" encoding="UTF-8"?>
```
zsiec: Perhaps this one could be `manifestFiltering4096Representation`

pkg/filters/dash_test.go
```diff
@@ -300,3 +301,145 @@ func TestDASHFilter_FilterManifest_filterStreams(t *testing.T) {
...
+      <Representation bandwidth="2048" codecs="avc" id="0"></Representation>
+    </AdaptationSet>
+    <AdaptationSet id="1" lang="en" contentType="audio">
+      <Representation bandwidth="256" codecs="ac-3" id="0"></Representation>
+    </AdaptationSet>
+  </Period>
+</MPD>
+`
+
+	filteredManifestWithOneAS := `<?xml version="1.0" encoding="UTF-8"?>
```
zsiec: Perhaps this one could be `manifestFiltering2048And4096Representations`

pkg/filters/dash.go
zsiec: Edit: I added a suggestion in this PR that supersedes this comment: https://github.com/cbsinteractive/bakery/pull/22

pkg/filters/dash.go
sam-eng: thanks for pointing that out! I've made the changes you suggested and if you could give them a quick look-over, that would be great

Issue #21, opened 2020-02-03
bakery: unit test for hls bandwidth filter


krystalmejia24:
We have no test for our hls filter. Currently we only support bandwidth filter on hls.

PR #22, opened 2020-02-05
Unit Tests for HLS Bandwidth Filter

sam-eng:
refs #21

### DIFF
--- a/pkg/filters/hls.go
+++ b/pkg/filters/hls.go
@@ -56,7 +56,6 @@ func (h *HLSFilter) FilterManifest(filters *parsers.MediaFilters) (string, error
 }
 
 func (h *HLSFilter) validateVariants(filters *parsers.MediaFilters, v *m3u8.Variant) bool {
-
 	if filters.DefinesBitrateFilter() {
 		if !(h.validateBandwidthVariant(filters.MinBitrate, filters.MaxBitrate, v)) {
 			return false
--- a/pkg/filters/hls.go
+++ b/pkg/filters/hls.go
@@ -2,7 +2,6 @@ package filters
 
 import (
 	"errors"
-	"fmt"
 	"math"
 	"path/filepath"
 	"strings"
@@ -60,7 +59,7 @@ func (h *HLSFilter) FilterManifest(filters *parsers.MediaFilters) (string, error
 func (h *HLSFilter) validateVariants(filters *parsers.MediaFilters, v *m3u8.Variant) bool {
 
 	if bandwidthInRange(filters.MinBitrate, filters.MaxBitrate) {
-		fmt.Println("Running Bandwidth Filter...")
+
 		if !(h.validateBandwidthVariant(filters.MinBitrate, filters.MaxBitrate, v)) {
 			return false
 		}
@@ -93,7 +92,7 @@ func bandwidthInRange(minBandwidth int, maxBandwidth int) bool {
 		(minBandwidth < maxBandwidth) &&
 		!(minBandwidth == 0 && maxBandwidth == math.MaxInt32) {
 		return true
-	} else {
-		return false
 	}
+
+	return false
 }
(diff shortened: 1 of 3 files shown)


pkg/filters/hls.go
```diff
@@ -47,17 +48,29 @@ func (h *HLSFilter) FilterManifest(filters *parsers.MediaFilters) (string, error
...
 	}
 
 	return filteredManifest.String(), nil
 }
 
-func (h *HLSFilter) validateVariant(filters *parsers.MediaFilters, v *m3u8.Variant) bool {
+func (h *HLSFilter) validateVariants(filters *parsers.MediaFilters, v *m3u8.Variant) bool {
+
+	if bandwidthInRange(filters.MinBitrate, filters.MaxBitrate) {
+
```
flavioribeiro:
```suggestion
```

pkg/filters/hls.go
```diff
@@ -47,17 +48,29 @@ func (h *HLSFilter) FilterManifest(filters *parsers.MediaFilters) (string, error
...
 			filteredManifest.Append(normalizedVariant.URI, normalizedVariant.Chunklist, normalizedVariant.VariantParams)
 		}
 	}
 
 	return filteredManifest.String(), nil
 }
 
-func (h *HLSFilter) validateVariant(filters *parsers.MediaFilters, v *m3u8.Variant) bool {
+func (h *HLSFilter) validateVariants(filters *parsers.MediaFilters, v *m3u8.Variant) bool {
+
```
flavioribeiro:
```suggestion
```

pkg/filters/hls.go
```diff
@@ -73,3 +86,13 @@ func (h *HLSFilter) normalizeVariant(v *m3u8.Variant, absoluteURL string) *m3u8.
...
 }
+
+func bandwidthInRange(minBandwidth int, maxBandwidth int) bool {
+	if (minBandwidth >= 0 && maxBandwidth <= math.MaxInt32) &&
+		(minBandwidth < maxBandwidth) &&
+		!(minBandwidth == 0 && maxBandwidth == math.MaxInt32) {
+		return true
+	}
+
+	return false
```
flavioribeiro:
```suggestion
	return (minBandwidth >= 0 && maxBandwidth <= math.MaxInt32) &&
		(minBandwidth < maxBandwidth) &&
		!(minBandwidth == 0 && maxBandwidth == math.MaxInt32)
```

flavioribeiro:
Tests are looking good! I just added a few comments.

pkg/filters/hls.go
```diff
@@ -73,3 +86,13 @@ func (h *HLSFilter) normalizeVariant(v *m3u8.Variant, absoluteURL string) *m3u8.
 
 	return v
 }
+
+func bandwidthInRange(minBandwidth int, maxBandwidth int) bool {
```
zsiec: I see we're duplicating this logic from the DASH filter. Another option would be to add a method to the `*parsers.MediaFilters` receiver, something like this:

```go
func (f *MediaFilters) definesBitrateFilter() bool {
    return (minBandwidth >= 0 && maxBandwidth <= math.MaxInt32) &&
        (minBandwidth < maxBandwidth) &&
        !(minBandwidth == 0 && maxBandwidth == math.MaxInt32)
}
``` 
(this would go in https://github.com/cbsinteractive/bakery/blob/master/pkg/parsers/urlparser.go)

And you'd change Line 61 to something like:
```go
if filters.definesBitrateFilter() {
```
And you can use this same function over in the DASH filter code!

pkg/filters/hls.go
sam-eng: thanks for pointing these out! I've made the changes you suggested

pkg/filters/hls.go
sam-eng: Thanks for the suggestions, I've made the changes--do you think you could take another quick look and make sure it looks good?

zsiec: The change looks good 👍 approved!

Issue #24, opened 2020-02-10
bakery: Filter audio and video codecs for hls


krystalmejia24:
refs #19 

We want the same filter in place for hls.

PR #30, opened 2020-02-19
HLS Audio, Video, and Caption Codec Filter

sam-eng:
refs #24

### DIFF
--- a/pkg/filters/hls.go
+++ b/pkg/filters/hls.go
@@ -104,8 +104,7 @@ func (h *HLSFilter) validateVariants(filters *parsers.MediaFilters, v *m3u8.Vari
 // Returns true if the given variant (variantCodecs) should be allowed through the filter for supportedCodecs of filterType
 func validateVariantCodecs(filterType ContentType, variantCodecs []string, supportedCodecs map[string]struct{}) bool {
 	var matchFilterType func(string) bool
-	typeInVariant := 0
-	matchInVariant := 0
+	var typeInVariant, matchInVariant int
 
 	switch {
 	case filterType == audioContentType:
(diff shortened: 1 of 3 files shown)


pkg/filters/hls.go
```diff
@@ -55,16 +55,81 @@ func (h *HLSFilter) FilterManifest(filters *parsers.MediaFilters) (string, error
...
+	}
+
 	return true
 }
 
+// Returns true if the given variant (variantCodecs) should be allowed through the filter for supportedCodecs of filterType
+func validateVariantCodecs(filterType ContentType, variantCodecs []string, supportedCodecs map[string]struct{}) bool {
+	var matchFilterType func(string) bool
+	typeInVariant := 0
+	matchInVariant := 0
```
zsiec: because ints have a default value of 0, I think you can replace this with:

```suggestion
var typeInVariant, matchInVariant int

File: pkg/filters/hls.go [hunk 2]
```diff
@@ -55,16 +55,81 @@ func (h *HLSFilter) FilterManifest(filters *parsers.MediaFilters) (string, error
...
+
+	switch {
+	case filterType == audioContentType:
+		matchFilterType = isAudioCodec
+	case filterType == videoContentType:
+		matchFilterType = isVideoCodec
+	case filterType == captionContentType:
+		matchFilterType = isCaptionCodec
```
zsiec: We may want to add a default case here where we return an error if the filter type is not recognized, otherwise we might run into a panic on line 120 if the func is nil.

Another option could be to do something like:

```go
matchFns := map[ContentType]func(string)bool{
    audioContentType:   isAudioCodec,
    videoContentType:   isVideoCodec,
    captionContentType: isCaptionCodec
}
```

then you could do something like:

```go
matchFilterType, found := matchFns[filterType]
// handle if 'found' is false by returning an error or default value
```

pkg/filters/hls.go
```diff
@@ -83,3 +148,23 @@ func (h *HLSFilter) normalizeVariant(v *m3u8.Variant, absoluteURL string) *m3u8.
 
 	return v
 }
+
+// Returns true if given codec is an audio codec (mp4a, ec-3, or ac-3)
+func isAudioCodec(codec string) bool {
+	return (ValidCodecs(codec, CodecFilterID("mp4a")) ||
```
zsiec: Perhaps we can export/add to the values in this package https://github.com/cbsinteractive/bakery/blob/7ff0bdae4446c0d6dae685b57d6c94805580875e/pkg/filters/filter.go#L26...L31

That way, they're defined in one place and can be used everywhere else

pkg/filters/hls_test.go
```diff
@@ -112,3 +112,805 @@ http://existing.base/uri/link_2.m3u8
...
+			}
+
+			if g, e := manifest, tt.expectManifestContent; g != e {
+				t.Errorf("FilterManifest() wrong manifest returned)\ngot %v\nexpected: %v\ndiff: %v", g, e,
+					cmp.Diff(g, e))
+			}
+
+		})
+	}
+}
```
zsiec: Impressive comprehensive testing! :)

Issue #25, opened 2020-02-11
Absolute URLs in HLS manifests


keithrichardscbs:
When I set the BAKERY_ORIGIN_HOST to a carrier (propeller auto-captions enabled channel) url, the urls to each of the media playlists are absolute urls. These urls are getting clobbered by Bakery.

BAKERY_ORIGIN_HOST=https://devcarrier-hls.amlg.io

> C02Z651LLVDQMBP:propeller krich0930$ http 'https://fifb9bsjd6.execute-api.us-east-1.amazonaws.com/dev/b(300000,1000000)/captions/bcbeb411-8efd-4c28-92a8-e210f7e29fc3/master.m3u8'

> #EXTM3U
> #EXT-X-VERSION:4
> #EXT-X-MEDIA:TYPE=CLOSED-CAPTIONS,GROUP-ID="CC",NAME="ENGLISH",DEFAULT=NO,LANGUAGE="ENG",URI="https://devcarrier-hls.amlg.io/captions/bcbeb411-8efd-4c28-92a8-e210f7e29fc3/"
> #EXT-X-MEDIA:TYPE=SUBTITLES,GROUP-ID="subs",NAME="English",DEFAULT=YES,AUTOSELECT=YES,LANGUAGE="en",FORCED="NO",URI="https://devcarrier-hls.amlg.io/captions/bcbeb411-8efd-4c28-92a8-e210f7e29fc3/https://devcarrier-hls.amlg.io/captions/bcbeb411-8efd-4c28-92a8-e210f7e29fc3/en.m3u8"
> #EXT-X-STREAM-INF:PROGRAM-ID=0,BANDWIDTH=356400,AVERAGE-BANDWIDTH=345400,CODECS="avc1.640015,mp4a.40.2",RESOLUTION=400x244,CLOSED-CAPTIONS="CC",SUBTITLES="subs",FRAME-RATE=29.000
> https://devcarrier-hls.amlg.io/captions/bcbeb411-8efd-4c28-92a8-e210f7e29fc3/https://cbss64eb-cbss64eb-ms-dev.global.ssl.fastly.net/cbssce1b/master/cbssce1b_1.m3u8
[...]

PR #32, opened 2020-02-26
Absolute URL Fix in HLS

sam-eng:
ref #25

### DIFF
--- a/pkg/filters/hls.go
+++ b/pkg/filters/hls.go
@@ -154,7 +154,6 @@ func (h *HLSFilter) validateBandwidthVariant(minBitrate int, maxBitrate int, v *
 }
 
 func (h *HLSFilter) normalizeVariant(v *m3u8.Variant, absolute url.URL) (*m3u8.Variant, error) {
-
 	for _, a := range v.VariantParams.Alternatives {
 		aUrl, aErr := combinedIfRelative(a.URI, absolute)
 		if aErr != nil {
@@ -177,7 +176,7 @@ func combinedIfRelative(uri string, absolute url.URL) (string, error) {
 	}
 	relative, err := isRelative(uri)
 	if err != nil {
-		return uri, err // subject to change, confirm with someone if this is right action to take
+		return uri, err
 	}
 	if relative {
 		combined, err := absolute.Parse(uri)
--- a/pkg/filters/hls.go
+++ b/pkg/filters/hls.go
@@ -60,6 +60,11 @@ func (h *HLSFilter) FilterManifest(filters *parsers.MediaFilters) (string, error
 		if err != nil {
 			return "", err
 		}
+		validatedFilters, err := h.validateVariants(filters, normalizedVariant)
+		if err != nil {
+			return "", err
+		}
+
 		if validatedFilters {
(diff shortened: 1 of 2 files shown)


pkg/filters/hls.go
```diff
@@ -145,14 +153,48 @@ func (h *HLSFilter) validateBandwidthVariant(minBitrate int, maxBitrate int, v *
 	return true
 }
 
-func (h *HLSFilter) normalizeVariant(v *m3u8.Variant, absoluteURL string) *m3u8.Variant {
+func (h *HLSFilter) normalizeVariant(v *m3u8.Variant, absolute url.URL) (*m3u8.Variant, error) {
+
```
zsiec: small nit: we can remove this newline

pkg/filters/hls.go
```diff
@@ -145,14 +153,48 @@ func (h *HLSFilter) validateBandwidthVariant(minBitrate int, maxBitrate int, v *
...
+}
 
-	return v
+func combinedIfRelative(uri string, absolute url.URL) (string, error) {
+	if len(uri) == 0 {
+		return uri, nil
+	}
+	relative, err := isRelative(uri)
+	if err != nil {
+		return uri, err // subject to change, confirm with someone if this is right action to take
```
zsiec: I agree that if we can't parse the url, an error makes sense!

pkg/filters/hls_test.go
```diff
@@ -914,3 +916,183 @@ http://existing.base/uri/link_13.m3u8
...
+
+	manifestWithIllegalVariantURLs := `#EXTM3U
+#EXT-X-VERSION:4
+#EXT-X-MEDIA:TYPE=CLOSED-CAPTIONS,GROUP-ID="CC",NAME="ENGLISH",DEFAULT=NO,LANGUAGE="ENG"
+#EXT-X-MEDIA:TYPE=AUDIO,GROUP-ID="AU",NAME="ENGLISH",DEFAULT=NO,LANGUAGE="ENG",URI="\nillegal.mp3u8"
+#EXT-X-STREAM-INF:PROGRAM-ID=0,BANDWIDTH=1000,AVERAGE-BANDWIDTH=1000,AUDIO="AU",VIDEO="VID",CLOSED-CAPTIONS="CC"
+http://existi\ng.base/uri/link_1.m3u8
+`
+
+	tests := []struct {
```
zsiec: These tests look good! I'd suggest adding one with relative paths that reference a parent directory like `../../../some/path" just to make sure we're covering that case correctly.

pkg/filters/hls.go
```diff
@@ -145,14 +153,48 @@ func (h *HLSFilter) validateBandwidthVariant(minBitrate int, maxBitrate int, v *
...
+			return uri, err
+		}
+		return combined.String(), err
+	}
+	return uri, nil
+}
+
+func isRelative(urlStr string) (bool, error) {
+	u, e := url.Parse(urlStr)
+	if e == nil {
```
zsiec: For consistency, I think it might be clearer to do a `!= nil` check here and keep the happy-path from nesting:

```go
...
if e != nil {
  return false, e
}

return !u.IsAbs(), nil
```

PR #42, opened 2020-03-06
Status codes

krystalmejia24:


pkg/handlers/handler.go
```diff
@@ -68,6 +68,12 @@ func fetchManifest(c config.Config, manifestURL string) (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("fetching manifest: %w", err)
 	}
+
+	if sc := resp.StatusCode; sc/100 > 3 {
+		fmt.Println(sc)
```
zsiec: I think we can remove this log line as long as we're logging the full error further up the call stack, otherwise this is great!

krystalmejia24: LMFAO my debug log haha

krystalmejia24:
Thanks @zsiec and @as ! I updated the order. good call that would have been a pain to troubleshoot.!

zsiec:
good catch @as !

Issue #35, opened 2020-02-26
bakery: ability to re-label audio description tags


flavioribeiro:
We'd like to be able to replace audio description tags for roku.

@jigneshdhruv can you paste here some examples and the diff between normal and roku manifests?

krystalmejia24:
currently being tested by Player team.

krystalmejia24:
first iteration of this was done by targeting the roku accessibility issue. if we need any more issues around audio description, lets file another ticket.

PR #44, opened 2020-03-09
overwrite role value when accessibility element is present

krystalmejia24:
This references issue #35 

We have an issue where roku is not able to grab descriptive audio for accessibility controls. Roku has a fix in their roadmap for 2020 but for now we need to relabel the `Role` element.

The idea is we pass a filter `role(description)` to enable the rewrite. This filter will look for all adaptation sets that carry the accessibility element with the following `schemeIdUri="urn:tva:metadata:cs:AudioPurposeCS:2007"`and rewrite the value of its role from `"alternate"` to `"description"`

All access team will run some tests in DEV so they can play with this feature before we promote to production.


pkg/filters/dash.go
```diff
@@ -236,6 +240,19 @@ func (d *DASHFilter) filterBandwidth(filters *parsers.MediaFilters, manifest *mp
 	}
 }
 
+//
+func (d *DASHFilter) updateRoleDescription(filters *parsers.MediaFilters, manifest *mpd.MPD) {
+	for _, period := range manifest.Periods {
+		for _, as := range period.AdaptationSets {
+			for i, accessibility := range as.AccessibilityElems {
+				if *accessibility.SchemeIdUri == "urn:tva:metadata:cs:AudioPurposeCS:2007" {
```
zsiec: for safety, I think we'll want to check to make sure `accessibility` is not `nil` before dereferencing here.

pkg/filters/dash.go
```diff
@@ -236,6 +240,19 @@ func (d *DASHFilter) filterBandwidth(filters *parsers.MediaFilters, manifest *mp
 	}
 }
 
+//
```
zsiec: small nil: empty comment

pkg/filters/dash_test.go
```diff
@@ -798,3 +798,91 @@ func TestDASHFilter_FilterManifest_bitrate(t *testing.T) {
...
+  <Period>
+    <AdaptationSet id="7357" lang="en" contentType="audio">
+      <Role schemeIdUri="urn:mpeg:dash:role:2011" value="alternate"></Role>
+      <Representation bandwidth="256" codecs="ac-3" id="1"></Representation>
+    </AdaptationSet>
+  </Period>
+</MPD>
+`
+
+	manifestWithOverwrittenRoleValue := `<?xml version="1.0" encoding="UTF-8"?>
```
zsiec: just a suggestion, this could be a func to make the tests more readable:

```go
func manifestWithOverwrittenRoleValue(v string) string {
  ...
}

zsiec:
I'm not sure the filter as defined is clear enough for me to understand what it does without further digging. I'd be worried about using a generic name like `role` as the filter here. Unless we have an good understanding on how this filter will be re-used for other needs in the future, How do you feel about creating a one-off filter without parameters, something like `transformDescAudioRole` as we're only accepting a single value in this code anyway?

krystalmejia24:
So after a convo with @zsiec about the clarity of these one-off functions that need to be executed for various devices that aren't up to spec. I came up with a small scheme to plug in these functions and execute from the url. 

in essence if you supply this as part of the path `[plugin1,plugin2,plugin2]` the value of each will equal the key to a map that will hold all of these functions definitions for execution. 

For now I am naming the key and function name both `dvsRoleOverride`. 

Let me know if any of this is super unclear or if you think it needs more tlc.

PR #45, opened 2020-03-10
Propeller redirect

krystalmejia24:


pkg/origin/propeller.go
```diff
@@ -0,0 +1,60 @@
...
+		URL:       propellerURL,
+		OrgID:     orgID,
+		ChannelID: channelID,
+	}, nil
+}
+
+func getPropellerChannelURL(host string, orgID string, channelID string) (string, error) {
+	pURL, err := url.Parse(host)
+	if err != nil {
+		fmt.Println(err)
```
zsiec: I think we can remove this line

krystalmejia24: nice catch. I removed it in the branch im working on for trimming.

Issue #34, opened 2020-02-26
bakery: implement video/audio bitrate filters 


flavioribeiro:
We need a way to target audio/video tracks on the bitrate. filter.

PR #48, opened 2020-03-16
Nested Audio/Video and Bitrate Filters for HLS and DASH

sam-eng:
ref #34

writing documentation before I merge

### DIFF
--- a/docs/_posts/2020-03-05-quick-start.md
+++ b/docs/_posts/2020-03-05-quick-start.md
@@ -87,11 +87,11 @@ Following the examples above you can start applying filters like so:
     <br>You can nest codec and bitrate filters within audio and video filters. For example, if you wanted to remove AAC (MPEG-4) audio and filter the results within 500 Kbps and 1MB, you could make the following request to Bakery:
 
     ```
-    http://bakery.dev.cbsivideo.com/a(codecs(mp4a),b(500,1000))/star_trek_discovery/S01/E01.m3u8
+    http://bakery.dev.cbsivideo.com/a(co(mp4a),b(500,1000))/star_trek_discovery/S01/E01.m3u8
     ```
     for a Propeller channel:
     ```
-    http://bakery.dev.cbsivideo.com/a(codecs(mp4a),b(500,1000))/propeller/<org-id>/<channel-id>.m3u8
+    http://bakery.dev.cbsivideo.com/a(co(mp4a),b(500,1000))/propeller/<org-id>/<channel-id>.m3u8
     ```
 
 For more specific details and usage examples on specific filters and the values accepted by each, check out our documentation for filters <a href="/filters">here</a>!
--- a/docs/_posts/2020-03-05-quick-start.md
+++ b/docs/_posts/2020-03-05-quick-start.md
@@ -84,14 +84,14 @@ Following the examples above you can start applying filters like so:
     ```
 
 4. **Nested Filters**
-    <br>You can nest codec and bitrate filters within audio and video filters. For example, if you wanted to remove AAC (MPEG-4) audio and filter the results within 500 Kbps and 1Mbps, you could make the following request to Bakery:
+    <br>You can nest codec and bitrate filters within audio and video filters. For example, if you wanted to remove AAC (MPEG-4) audio and filter the results within 500 Kbps and 1MB, you could make the following request to Bakery:
 
     ```
-    http://bakery.dev.cbsivideo.com/a(codecs(mp4a),b(500000,1000000))/star_trek_discovery/S01/E01.m3u8
+    http://bakery.dev.cbsivideo.com/a(codecs(mp4a),b(500,1000))/star_trek_discovery/S01/E01.m3u8
     ```
     for a Propeller channel:
     ```
-    http://bakery.dev.cbsivideo.com/a(codecs(mp4a),b(500000,1000000))/propeller/<org-id>/<channel-id>.m3u8
+    http://bakery.dev.cbsivideo.com/a(codecs(mp4a),b(500,1000))/propeller/<org-id>/<channel-id>.m3u8
     ```
 
 For more specific details and usage examples on specific filters and the values accepted by each, check out our documentation for filters <a href="/filters">here</a>!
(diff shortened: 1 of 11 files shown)


pkg/filters/dash.go
```diff
@@ -209,22 +209,48 @@ func matchCodec(codec string, ct ContentType, supportedCodecs map[string]struct{
 func (d *DASHFilter) filterBandwidth(filters *parsers.MediaFilters, manifest *mpd.MPD) {
 	for _, period := range manifest.Periods {
 		var filteredAdaptationSets []*mpd.AdaptationSet
-
 		for _, as := range period.AdaptationSets {
 			var filteredRepresentations []*mpd.Representation
+			if as.ContentType != nil {
```
zsiec: I find that deep nesting can get hard to follow, you can remove one level by doing:

```go
if as.ContentType != nil {
    continue
}
...
```

From Mr Cheney:

>Return early rather than nesting deeply
>Every time you indent you add another precondition to the programmer’s stack consuming one of the 7 ±2 slots in their short term memory. Avoid control flow that requires deep indentation. Rather than nesting deeply, keep the success path to the left using guard clauses.

https://the-zen-of-go.netlify.com/

sam-eng: Thanks for the suggestion! I didn't know that this could have that sort of effect on the stack.

docs/filters/nested-filters.md
```diff
@@ -0,0 +1,63 @@
...
+### Multiple Filters:
+To use multiple filters, separated with `/` with no space between filters. You can use nested filters in conjunction with the general filters, such as the bandwidth filter.
+
+    // Removes AVC video, MPEG-4 audio, audio not in range of 500Kbps to 1Mbps
+    $ http http://bakery.dev.cbsivideo.com/v(codec(avc))/a(codec(mp4a),b(500000,1000000))/star_trek_discovery/S01/E01.m3u8
+
+    // Removes AVC video, MPEG-4 audio, and everything not in range of 500Kbps to 1Mbps
+    $ http http://bakery.dev.cbsivideo.com/v(codec(avc))/a(codec(mp4a))/b(500000,1000000)/star_trek_discovery/S01/E01.m3u8
+
+    // Removes AVC video, all video not in range 750Kbps to 1Mbps, MPEG-4 audio, and non-video not in range of 500Kbps to 1Mbps
```
krystalmejia24: Lets expand this a bit. We should highlight how the nested bandwidth filter will interact if you pass an overall bandwidth filter.

docs/filters/nested-filters.md
```diff
@@ -0,0 +1,63 @@
...
+## Supported Values
+
+| content types | example                     |
+|:-------------:|:---------------------------:|
+| audio         | a(codec(ac-3),b(0,1000000)) |
+| video         | v(codec(avc),b(0,1000000))  |
+
+| subfilters | example         |
+|:----------:|:---------------:|
+| codec      | a(codec(ac-3))  |
```
krystalmejia24: lets change the key for codec

`a(co(ac-3))`

I would say lets not use c since it signals "captions"

krystalmejia24: Lets go into the individual audio and video codec filters doc, and make a note that this is the default behavior of `v` and `a` but you can also add multiple filters to the audio and video key. and then link to this page. Does that make sense.

krystalmejia24: lets also add nested filters to the quick start guide.

pkg/parsers/urlparser.go
```diff
@@ -63,8 +74,8 @@ type Trim struct {
 
 // MediaFilters is a struct that carry all the information passed via url
 type MediaFilters struct {
-	Videos            []VideoType       `json:",omitempty"`
-	Audios            []AudioType       `json:",omitempty"`
+	VideoFilters      Subfilters        `json:",omitempty"`
```
krystalmejia24: How about changing the type of this to `Nestedfilters` ? I think its easier to follow along if we follow the nomenclature  in docs and code base.

docs/filters/bandwidth.md
```diff
@@ -15,7 +15,7 @@ yes | yes  |
 
 ## Supported Values
 
-| values (Kbps) | example   |
+| values (bps) | example   |
```
krystalmejia24: We will want to change this back to Kbps and update the values before.

docs/filters/nested-filters.md
sam-eng: done!

docs/_posts/2020-03-05-quick-start.md
```diff
@@ -83,6 +83,17 @@ Following the examples above you can start applying filters like so:
...
 
+4. **Nested Filters**
+    <br>You can nest codec and bitrate filters within audio and video filters. For example, if you wanted to remove AAC (MPEG-4) audio and filter the results within 500 Kbps and 1MB, you could make the following request to Bakery:
+
+    ```
+    http://bakery.dev.cbsivideo.com/a(codecs(mp4a),b(500,1000))/star_trek_discovery/S01/E01.m3u8
+    ```
+    for a Propeller channel:
+    ```
+    http://bakery.dev.cbsivideo.com/a(codecs(mp4a),b(500,1000))/propeller/<org-id>/<channel-id>.m3u8
```
krystalmejia24: @sam-eng we need to update this example to `co` and not `codec` for the nested filter key.

docs/filters/codec.md
```diff
@@ -6,6 +6,8 @@ nav_order: 1
 
 # Codec
 Values in this filter define a whitelist of the audio and video codecs you want to **EXCLUDE** in the modifed manifest. Passing an empty value for either video or audio will return all audio and video codecs available in the manifest.
+<br>
+By default, the audio and video filter keys take codecs, but this isn't the only way to use them. You can nest filters to filter more than just codecs, as detailed on the <a href="nested-filters.html">nested filters</a> page.
```
krystalmejia24: LOVE THIS!

docs/filters/nested-filters.md
```diff
@@ -0,0 +1,67 @@
...
+| codec      | a(co(ac-3))  |
+| bandwidth  | v(b(0,1000)) |
+
+
+## Usage Example
+### Single Nested Filter:
+
+    // Removes MPEG-4 audio
+    $ http http://bakery.dev.cbsivideo.com/a(codec(mp4a))/star_trek_discovery/S01/E01.m3u8
+
```
krystalmejia24: Will also need to make changes here from `codec` to `co` nested key

krystalmejia24:
Looks great! There is an outdated key in the docs. But should be good to merge once upated!

Issue #37, opened 2020-03-02
hls: trim live hls stream to specified start/end and return VOD playlists for variants


zsiec:
For the Big Brother usecase, we need to be able to harvest up to 24 hours of video off the live streams every day. Mediahub will take care of generating these VODs, but expects a correctly trimmed VOD HLS stream as input. Bakery will be responsible for generating this VOD HLS stream for Mediahub based on the config defined by Propeller.

The Propeller team is setting up a MediaStore container where they will push clear streams from MediaLive. The live stream will retain segments and return variant playlists with up to 48 hours of content.

In order to fetch content from only the last 24h, requests will be made to bakery with a start and end time specifying what parts of the manifest to include in the modified stream.

Assuming:
1) We define a filter named `trim` where the syntax is `t(start,end)` (initial thought is to use 64-bit UNIX time for these values)
2) Propeller exposes the clear stream url in their channel payload: example http://propeller-api.cbsivideo.com/v1/organization/gamesg041/channel/pokemf87h
3) The stream url contains valid `#EXT-X-PROGRAM-DATE-TIME` tags with content inside the range requested in the `trim` filter

The initial proposal is to interpret a master manifest URL like:
https://bakery.cbsi.video/t(start,end)/propeller/gamesg041/channel/pokemf87h.m3u8

When Bakery sees this request, it will:
1) Request the live stream url from Propeller's API
2) Re-write the media manifest URLs to point to bakery, passing through the trimming config, generating media manifest urls like: `https://bakery.cbsi.video/t(start,end)/propeller/gamesg041/channel/pokemf87h/<base64-encoded-original-media-manifest-url>.m3u8`

When Bakery sees this media manifest, it will:
1) Parse out the base64 encoded url, decode and fetch manifest contents
2) Trim out any segments outside the defined `trim` range
3) Append an `#EXT-X-ENDLIST` to the end of the manifest
4) Rewrite segment urls if they are relative, to ensure they are coming from the original source

The stream propeller exposes for harvesting these VODs will require an auth mechanism, (likely a custom header) Bakery will need to figure out how this will be configured and ensure the auth header is sent on all requests to the the source stream.

krystalmejia24:
@zsiec @allisonnicoledeal do you have a stream I can test with? I can create a test clip in propeller im mostly trying to wrap my head around the playlist type? Is it live with a sliding window or is it VOD and it will have the full list of segments advertised in the manifest?

krystalmejia24:
[...]

PR #49, opened 2020-03-16
Trim hls

krystalmejia24:
Refs issue #37 

There are some assumptions written in the issue that I'd like to clarify. May need to make some changes to propeller for accessing that bakery URL. CC @allisonnicoledeal 

When you request a propeller stream via bakery in the form of 

`https://bakery.dev.cbsivideo.com/t(1583971200,1584057600)/propeller/orgID/channelID.m3u8`

Bakery will fetch the Propeller stream and rewrite the variant URLs to point to bakery in the form of 

`https://bakery.dev.cbsivideo.com/t(1583971200,1584057600)/<base64-encoded-original-media-manifest-url>.m3u8`

Currently, there is a dependency on PDT for every segments. Since propeller is adding PDT to each segment we will be fine with trimming any streams with a Propeller origin.


pkg/filters/hls.go
```diff
@@ -174,6 +186,18 @@ func (h *HLSFilter) normalizeVariant(v *m3u8.Variant, absolute url.URL) (*m3u8.V
 	return v, nil
 }
 
+func (h *HLSFilter) normalizeTrimmedVariant(filters *parsers.MediaFilters, uri string) (string, error) {
+	encoded := base64.StdEncoding.EncodeToString([]byte(uri))
+	start := filters.Trim.Start
+	end := filters.Trim.End
+	u, err := url.Parse(uri)
```
zsiec: Will this work for master manifests with relative paths to media manifests?

pkg/filters/hls.go
```diff
@@ -174,6 +186,18 @@ func (h *HLSFilter) normalizeVariant(v *m3u8.Variant, absolute url.URL) (*m3u8.V
...
+	if err != nil {
+		return "", err
+	}
+
+	return fmt.Sprintf("%v://%v/t(%v,%v)/%v.m3u8", u.Scheme, h.config.Hostname, start, end, encoded), nil
```
zsiec: Curious if this works for master manifests with relative paths to media manifests?

pkg/filters/hls.go
```diff
@@ -200,22 +224,56 @@ func isRelative(urlStr string) (bool, error) {
...
 
-// Returns true if given codec is an audio codec (mp4a, ec-3, or ac-3)
-func isAudioCodec(codec string) bool {
-	return (ValidCodecs(codec, aacCodec) ||
-		ValidCodecs(codec, ec3Codec) ||
-		ValidCodecs(codec, ac3Codec))
+// FilterRenditionManifest will be responsible for filtering the manifest
+// according  to the MediaFilters
+func (h *HLSFilter) filterRenditionManifest(filters *parsers.MediaFilters, m *m3u8.MediaPlaylist) (string, error) {
+	filteredPlaylist, _ := m3u8.NewMediaPlaylist(uint(len(m.Segments)), uint(len(m.Segments)))
```
zsiec: I think we should handle this error by early returning it to the caller.

pkg/filters/hls.go
```diff
@@ -200,22 +224,56 @@ func isRelative(urlStr string) (bool, error) {
...
+	seqID := 0
```
zsiec: When declaring a variable initialized with its default value, you could use `var seqID int` instead of the `:=` assignment.

pkg/filters/hls.go
```diff
@@ -200,22 +224,56 @@ func isRelative(urlStr string) (bool, error) {
...
+
+	for _, segment := range m.Segments {
+		if segment != nil {
```
zsiec: I find that deep nesting can get hard to follow, you can remove one level by doing:

```go
if segment == nil {
    continue
}

File: pkg/filters/hls.go [hunk 6]
```diff
@@ -200,22 +224,56 @@ func isRelative(urlStr string) (bool, error) {
...
+					return "", fmt.Errorf("trimming segments: %w", err)
+				}
+
+				if isRelativeURL {
+					absolute, err := getAbsoluteURLString(h.manifestURL)
+					if err != nil {
+						return "", fmt.Errorf("fetching absolute url: %w", err)
+					}
+
+					segment.URI = absolute + segment.URI
```
zsiec: You may want to check out the code Same wrote here: https://github.com/cbsinteractive/bakery/pull/32/files#diff-f8d027c872dd51f0e133c3c45b300fa9R173-R189 for a way to do this using the `url` package instead of concat-ing strings.

pkg/filters/hls.go
```diff
@@ -200,22 +224,56 @@ func isRelative(urlStr string) (bool, error) {
...
+				if err != nil {
+					return "", fmt.Errorf("trimming segments: %w", err)
+				}
+
+				seqID++
+			}
+		}
+	}
+
+	filteredPlaylist.SeqNo = uint64(seqID)
```
zsiec: If we're outputting VOD, shouldn't we keep this at 0?

pkg/filters/hls.go
zsiec: I see later in a test you've proven it does, ignore me!

pkg/filters/hls.go
```diff
@@ -174,6 +186,18 @@ func (h *HLSFilter) normalizeVariant(v *m3u8.Variant, absolute url.URL) (*m3u8.V
 	return v, nil
 }
 
+func (h *HLSFilter) normalizeTrimmedVariant(filters *parsers.MediaFilters, uri string) (string, error) {
+	encoded := base64.StdEncoding.EncodeToString([]byte(uri))
```
zsiec: Because we'll be using this value as a URL path element, I'd recommend we use `RawURLEncoding`: https://golang.org/src/encoding/base64/base64.go?s=6840:6898#L113

If we change this, we have to remember to update the encoding we use when we decode as well.

pkg/origin/origin.go
```diff
@@ -15,46 +15,79 @@ type Origin interface {
 }
 
 //Manifest struct holds Origin and Path of Manifest
+//Variant level manifests will be base64 encoded absolute path
 type Manifest struct {
-	Origin string
-	Path   string
+	Origin   string
+	Path     string
```
zsiec: Maybe this should be type `url.URL` so you don't need to cache things like `Absolute` in the struct. Functions like `GetPath` might get simpler as well.

pkg/origin/propeller.go
```diff
@@ -20,28 +22,54 @@ func (p *Propeller) GetPlaybackURL() string {
...
+	case "": //fetch master
+		propellerURL, err = getPropellerChannelURL(c.PropellerHost, orgID, channelID)
+		if err != nil {
+			return &Propeller{}, fmt.Errorf("fetching propeller channel: %w", err)
+		}
+	default:
+		propellerURL, err = decodeRenditionURL(renditionURL)
+		if err != nil {
+			return &Propeller{}, fmt.Errorf("fetching propeller rendition: %w", err)
+		}
```
zsiec: This feels a little awkward to have and empty string for `renditionURL` imply we need to run logic for a master manifest. Perhaps we could try different constructors? It could also be a hint that this struct is doing more than one thing and could use a separation of concerns.

pkg/parsers/urlparser.go
```diff
@@ -54,6 +54,12 @@ const (
 	ProtocolDASH Protocol = "dash"
 )
 
+// Trim is a struct that carries the start and end times to trim playlist
+type Trim struct {
+	Start int `json:",omitempty"`
+	End   int `json:",omitempty"`
```
zsiec: Maybe we should explicitly make these `int64` to avoid the casting you do in the filter logic?

pkg/parsers/urlparser.go
```diff
@@ -148,6 +155,17 @@ func URLParse(urlpath string) (string, *MediaFilters, error) {
...
 				mf.MaxBitrate, _ = strconv.Atoi(filters[1])
 			}
+		case "t":
+			var trim Trim
+			if filters[0] != "" {
+				trim.Start, _ = strconv.Atoi(filters[0])
+			}
+
+			if filters[1] != "" {
+				trim.End, _ = strconv.Atoi(filters[1])
```
zsiec: Shouldn't we handle these errors to be able to provide useful feedback when we've misconfigured the url?

pkg/filters/hls.go
```diff
@@ -200,22 +224,56 @@ func isRelative(urlStr string) (bool, error) {
...
-		ValidCodecs(codec, ac3Codec))
+// FilterRenditionManifest will be responsible for filtering the manifest
+// according  to the MediaFilters
+func (h *HLSFilter) filterRenditionManifest(filters *parsers.MediaFilters, m *m3u8.MediaPlaylist) (string, error) {
+	filteredPlaylist, _ := m3u8.NewMediaPlaylist(uint(len(m.Segments)), uint(len(m.Segments)))
+	seqID := 0
+
+	for _, segment := range m.Segments {
+		if segment != nil {
+			if inRange(int64(filters.Trim.Start), int64(filters.Trim.End), segment.ProgramDateTime.Unix()) {
```
zsiec: Would it make sense to return an error if the value for `segment.ProgramDateTime.Unix()` returned the default value for `time.Time` (signifying the PDT wasn't present)?

pkg/origin/origin.go
```diff
@@ -15,46 +15,79 @@ type Origin interface {
...
+		path = renditionURL
+	}
+
 	return NewManifest(c, path), nil
 }
 
 //NewManifest returns a new Origin struct
 func NewManifest(c config.Config, path string) *Manifest {
+	var absolute bool
+	if strings.Contains(path, "http") {
```
zsiec: If we don't decide to change the type for `path` to `url.URL` (which would be nice to be able to use `.IsAbs()`), we should be careful about using `strings.Contains` for checks like this. The string "$$$$$$$http$$$"  would pass this check. We could use `strings.HasPrefix` for some more safety. I think I've done things like `strings.HasPrefix(s, "http://") || strings.HasPrefix(s, "https://")` in the past.

pkg/filters/hls.go
krystalmejia24: correct!

pkg/filters/hls.go
krystalmejia24: fixed.

pkg/filters/hls.go
krystalmejia24: removed this since VOD playlist start at 0

pkg/filters/hls.go
krystalmejia24: dope i updated my code a bit to use that. thanks for pointing it out.

pkg/parsers/urlparser.go
krystalmejia24: done!

pkg/parsers/urlparser.go
krystalmejia24: good point. i think i got lazy bc i noticed we had no validation in there for bitrate filter conversions. 
Added something small to do both trim and bitrates along with the tests. Simple error check during the conversions from strings and making sure start/min values are not greater than end/max values

pkg/origin/propeller.go
krystalmejia24: Im no longer using Propeller struct to manage the encoded rendition urls. This was a first pass and I never cleaned up!

Essentially i move the logic to the Manifest struct. If path is absolute URL that means its the decoded rendition url.

pkg/origin/origin.go
krystalmejia24: GetPath was in there as a way for me to interface with the propeller struct (which only holds org id and channel id) so that I can get back the original path request `propeller/orgid/channel.m3u8` when writing the rendition urls in the master manifest with the schema `propeller/orgid/channel/base64.m3u8`

I realized when writing those urls that honoring that schema  was overkill since we shouldn't care if the rendition came from propeller or a different origin.

krystalmejia24: I've actually updating `Path` to `url.Url` because it reads much cleaner to me and if i need to extent this in the future i think having it as a url type will make it easier to manipulate.

pkg/origin/origin.go
krystalmejia24: i changed to `url.Url` as described above but I still am using the `path.IsAbs()` bc as you pointed out its better than looking for `http` in the string.

pkg/filters/hls.go
krystalmejia24: this one slipped me. good eye.

pkg/filters/hls.go
krystalmejia24: updated both encoding and decoding! thanks for this.

pkg/filters/hls.go
krystalmejia24: done! added a test for this. thanks bc we would have gotten to a situation where we serve an empty playlist for the play to try to serve.

pkg/filters/hls.go
```diff
@@ -200,22 +224,52 @@ func isRelative(urlStr string) (bool, error) {
...
+		return "", fmt.Errorf("filtering Rendition Manifest: %w", err)
+	}
+
+	for _, segment := range m.Segments {
+		if segment == nil {
+			continue
+		}
+
+		if segment.ProgramDateTime == (time.Time{}) {
+			return "", fmt.Errorf("Program Date Time not set on segments")
```
zsiec: tinyest of nits:
`Error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with punctuation, since they are usually printed following other context.`
https://github.com/golang/go/wiki/CodeReviewComments#error-strings

pkg/origin/origin.go
```diff
@@ -15,9 +17,10 @@ type Origin interface {
 }
 
 //Manifest struct holds Origin and Path of Manifest
+//Variant level manifests will be base64 encoded absolute path
 type Manifest struct {
 	Origin string
-	Path   string
+	Path   url.URL
```
zsiec: Maybe we should rename this to `URL` or `OriginalURL`?

krystalmejia24: went with URL for clarity

PR #53, opened 2020-03-24
remove aws lambda dependencies

flavioribeiro:
we no longer deploy bakery as a lambda. removing the dependencies and code for that.


krystalmejia24:
thanks @flavioribeiro !

Issue #52, opened 2020-03-17
Manage Propeller URLs in bakery


krystalmejia24:
There are a few different URLS floating around in propeller to handle all of our different integrations. For now we are exposing PlaybackURL but will need to reorganize priority of how thats done. Based on the convo with @flavioribeiro and @igorsobreira , lets handle URLs with the following Priority:

1) ads
2) captions
3) playbackURL

We also need to keep track of the work being done to support multiple [output groups ](https://github.com/cbsinteractive/propeller/issues/790) as that will lead to updating the url scheme to something in the form of:

 `bakery-hostname/propeller/org-id/channel-id/label.m3u8` 

where label will correspond to an output group.

PR #54, opened 2020-03-25
Propeller url

krystalmejia24:
refs #52 

This will handle authentication. The creds are deployed via env vars. To accommodate this change I moved the Propeller Client to be created on startup. I don't think dynamically creating a new client per request is smart anyway.

I've added an internal func for managing the URLs. In order of Ads, Captions, PlaybackURL.

will open a new issue for tracking the work being done by output groups.

### DIFF
--- a/docs/_posts/2020-03-05-quick-start.md
+++ b/docs/_posts/2020-03-05-quick-start.md
@@ -24,7 +24,7 @@ Then your `BAKERY_ORIGIN_HOST` was set to `http://streaming.cbsi.video` and your
 
 Bakery can act as a proxy to serve your Propeller channels!
 
-Propeller is the ViacomCBS live-streaming platform. Propeller can orchestrate and manages all of your cloud based resources needed for a live streaming environment. If you have a stream running out of Propeller, you can stream that channel via Bakery! 
+Propeller is the ViacomCBS live-streaming platform. Propeller can orchestrate and manage all of your cloud based resources needed for a live streaming environment. If you have a stream running out of Propeller, you can stream that channel via Bakery! 
 
 
 To play a Propeller channel you can make a Bakery request with the following scheme:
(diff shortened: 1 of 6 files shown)


docs/_posts/2020-03-05-quick-start.md
```diff
@@ -24,7 +24,7 @@ Then your `BAKERY_ORIGIN_HOST` was set to `http://streaming.cbsi.video` and your
 
 Bakery can act as a proxy to serve your Propeller channels!
 
-Propeller is a live-streaming platform that manages all of your cloud based resources. If you have a stream running out of Propeller, you can stream that channel via Bakery! 
+Propeller is the ViacomCBS live-streaming platform. Propeller can orchestrate and manages all of your cloud based resources needed for a live streaming environment. If you have a stream running out of Propeller, you can stream that channel via Bakery!
```
zsiec:
```suggestion
Propeller is the ViacomCBS live-streaming platform. Propeller can orchestrate and manage all of your cloud based resources needed for a live streaming environment. If you have a stream running out of Propeller, you can stream that channel via Bakery! 
```

pkg/origin/propeller.go
```diff
@@ -39,22 +38,28 @@ func NewPropeller(c config.Config, orgID string, channelID string) (*Propeller,
...
 	if err != nil {
-		return "", fmt.Errorf("parsing propeller host url: %w", err)
+		return "", fmt.Errorf("fetching channel from propeller: %w", err)
 	}
-	p := client.NewClient(pURL)
 
-	channel, err := p.GetChannel(orgID, channelID)
-	if err != nil {
-		return "", fmt.Errorf("fetching channel from propeller: %w", err)
+	return getURL(*channel)
```
zsiec: Curious why `GetChannel()` returns a pointer to a Channel requiring the dereference here?

krystalmejia24: Ahh so looking at the client below I had initiated response as a pointer rather than just referencing it during the request.  I just changed this as I was doing it on all of the requests. 

```// GetChannel returns a propeller Channel
func (c *Client) GetChannel(orgID string, channelID string) (*Channel, error) {
	result := &Channel{}
	path := fmt.Sprintf("/v1/organization/%v/channel/%v", orgID, channelID)

	err := c.get(path, result)
	if err != nil {
		return result, err
	}

	return result, nil
}```

PR #57, opened 2020-03-27
I-Frame & Language filters

krystalmejia24:
1. This PR also has some refactoring in the parsers package where i
    - In the parsers package I refactored the MediaFilters by creating a `Bitrate` struct that holds `Min` and `Max`. If no bitrate filters are set, then it will be set to nil during the filtering of the manifest and we no longer need to make checks for whether the boundaries were implicitly or explicitly set.
    - Added simple validation for range of int type filters. Both Trim and Bandwidth filters will now check for a positive range. 

2. I frame filtering for hls
    - simply pass it as a video type filter `v(i-frame)`

3. Language filtering
    - you can pass an overall lang filter on request `l(es,pt)` which will then remove captions AND audio languages
    - you can pass a nested lang filter to the audio or caption type `a(l(es,pt))` or `c(l(es,pt))` if you only needed to filter out one content type

writing up docs but will include in sep PR since I don't want those merged in until these make it into production.


krystalmejia24:
I've gone ahead and extended the actions workflows to include deployment triggers in this PR. I've created two more jobs, `deploy_dev` and `deploy_prod`.  Both deploy jobs require test job to pass. `deploy_dev` only runs when pushed to master. `deploy_prod` only runs when you push tags that match this `v\d+\.\d+\.\d+$` 

I've tested the dispatch event receiver in the bakery-deploy side by locally pushing dispatch_events and manually pushing tags in bakery to make sure it all works accordingly.  Its already in master on bakery-deploy (dispatch events only can be triggered from master). Check that out [here](https://github.com/cbsinteractive/bakery-deploy/blob/master/.github/workflows/deploy.yml)

Ive added status badges to this repo for both the build and the deploy.

PR #78, opened 2020-04-11
Bakery remove pkg / update mod

krystalmejia24:
I restructured and removed pkg. 
This work also includes pointing to the latest parser to support twitch tags.


krystalmejia24:
@as Thanks for the quickness!

PR #98, opened 2020-05-03
Add support for propeller outputs urls

igorsobreira:
Propeller added support for multiple outputs (https://github.com/cbsinteractive/propeller/pull/895)

Right now bakery support the following URLs:
```
/propeller/orgID/channelID.m3u8
/propeller/orgID/clip/clipID.m3u8
```
this change will add support for
```
/propeller/orgID/channelID/outputID.m3u8
```

Required change on propeller-go client: https://github.com/cbsinteractive/propeller-go/pull/4

**Note on tests**

The first commit in this PR adds some tests for `origin.Configure()` 
for propeller endpoints. The test starts a fake Propeller API with stubbed
responses allowing origin.NewPropeller() to interact with the API
and build a `origin.Propeller` object

For situations like this I usually prefer to mock the go client itself,
but Bajery used uses a reference to `*propeller.Client` which makes it
harder to replace for tests. We could refactor that into an interface,
if Bakery works with a `propeller.Client` interface then it's easy to swap
with a mock implementation for tests.

I didn't want to do any design changes before adding support for
propeller outputs urls. So this test helps increase coverage for
`origin.NewPropeller()`

### DIFF
--- a/origin/origin_test.go
+++ b/origin/origin_test.go
@@ -262,7 +262,10 @@ func configMockPropellerAPI() (cfg config.Config, teardown func()) {
 		}
 	}))
 
-	tsURL, _ := url.Parse(ts.URL)
+	tsURL, err := url.Parse(ts.URL)
+	if err != nil {
+		panic(fmt.Sprintf("go httptest server returned invalid url: %v (%v)", ts.URL, err))
+	}
 	cfg = config.Config{LogLevel: "panic"}
 	cfg.Propeller.Client.HostURL = tsURL
 
--- a/origin/origin_test.go
+++ b/origin/origin_test.go
@@ -172,47 +172,29 @@ func TestOrigin_Configure(t *testing.T) {
 		expectErr bool
 	}{
 		{
-			name:      "when origin is of type propeller in wrong format, return error",
-			path:      "/propeller/chanID.m3u8",
-			c:         config.Config{LogLevel: "panic"},
-			expected:  &Propeller{},
-			expectErr: true,
-		},
-		{
-			name:     "when origin path is of type propeller for channel manifest with simple playback_url",
-			path:     "/propeller/org123/ch-playback-url.m3u8",
+			name:     "when origin path is of type propeller for channel manifest",
+			path:     "/propeller/org123/channel-123.m3u8",
 			c:        cfg,
 			expected: &Propeller{URL: "http://cdn.com/ch.m3u8"},
 		},
 		{
-			name:     "when origin path is of type propeller for channel manifest with captions",
-			path:     "/propeller/org123/ch-captions-url.m3u8",
+			name:     "when origin path is of type propeller for channel output manifest",
+			path:     "/propeller/org123/channel-with-output/output-123.m3u8",
 			c:        cfg,
-			expected: &Propeller{URL: "http://captions.com/ch.m3u8"},
-		},
-		{
-			name:     "when origin path is of type propeller for channel manifest with ads and status running",
-			path:     "/propeller/org123/ch-ads-running.m3u8",
-			c:        cfg,
-			expected: &Propeller{URL: "http://ads.com/ch.m3u8"},
-		},
-		{
-			name:     "when origin path is of type propeller for channel manifest with ads and status not running",
-			path:     "/propeller/org123/ch-ads-not-running.m3u8",
-			c:        cfg,
-			expected: &Propeller{URL: "http://cdn.com/ch.m3u8"}, // use playback_url
-		},
-		{
-			name:     "when origin path is of type propeller for channel manifest when not found try clip archive",
-			path:     "/propeller/org123/ch-not-found.m3u8",
-			c:        cfg,
-			expected: &Propeller{URL: "http://cdn.com/ch-archive.m3u8"},
+			expected: &Propeller{URL: "http://cdn.com/output-123.m3u8"},
 		},
 		{
 			name:     "when origin path is of type propeller for clip manifest",
 			path:     "/propeller/org123/clip/clip-123.m3u8",
 			c:        cfg,
 			expected: &Propeller{URL: "http://cdn.com/clip-123.m3u8"},
+		},
+		{
+			name:      "when origin is of type propeller in wrong format, return error",
+			path:      "/propeller/chanID.m3u8",
+			c:         config.Config{LogLevel: "panic"},
+			expected:  &Propeller{},
+			expectErr: true,
 		},
 		{
 			name:     "when origin path is at root but not base64 encoded, return default origin type",
@@ -233,7 +215,7 @@ func TestOrigin_Configure(t *testing.T) {
 			got, err := Configure(tc.c, tc.path)
 
 			if err != nil && !tc.expectErr {
-				t.Errorf("Configure() didnt expect an error to be returned, got: %v", err)
+				t.Errorf("Configure() didnt expect an error to be returned, got: %q", err)
 				return
 			} else if err == nil && tc.expectErr {
 				t.Error("Configure() expected an error, got nil")
@@ -241,7 +223,7 @@ func TestOrigin_Configure(t *testing.T) {
 			}
 
 			if !cmp.Equal(got, tc.expected) {
-				t.Errorf("Wrong Origin returned\ngot %v\nexpected: %v\ndiff: %v",
+				t.Errorf("Wrong Origin returned\ngot %q\nexpected: %q\ndiff: %v",
 					got, tc.expected, cmp.Diff(got, tc.expected))
 			}
 
@@ -257,35 +239,17 @@ func TestOrigin_Configure(t *testing.T) {
 func configMockPropellerAPI() (cfg config.Config, teardown func()) {
 	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		switch r.URL.String() {
-		case "/v1/organization/org123/channel/ch-playback-url":
+		case "/v1/organization/org123/channel/channel-123":
 			fmt.Fprint(w, `{
 				"playback_url": "http://cdn.com/ch.m3u8"
 			}`)
-		case "/v1/organization/org123/channel/ch-captions-url":
+		case "/v1/organization/org123/channel/channel-with-output":
 			fmt.Fprint(w, `{
-				"auto_captions": true,
-				"playback_url_auto_captions": "http://captions.com/ch.m3u8"
-			}`)
-		case "/v1/organization/org123/channel/ch-ads-running":
-			fmt.Fprint(w, `{
-				"ads": true,
-				"status": "running",
-				"playback_url_ads": "http://ads.com/ch.m3u8"
-			}`)
-		case "/v1/organization/org123/channel/ch-ads-not-running":
-			fmt.Fprint(w, `{
-				"ads": true,
-				"status": "ready",
 				"playback_url": "http://cdn.com/ch.m3u8",
-				"playback_url_ads": "http://ads.com/ch.m3u8"
-			}`)
-		case "/v1/organization/org123/channel/ch-not-found":
-			w.WriteHeader(404)
-			fmt.Fprint(w, `{"message": "channel not found"}`)
-		case "/v1/organization/org123/clip/ch-not-found-archive":
-			fmt.Fprint(w, `{
-				"status": "created",
-				"url": "http://cdn.com/ch-archive.m3u8"
+				"outputs": [{
+					"id": "output-123",
+					"playback_url": "http://cdn.com/output-123.m3u8"
+				}]
 			}`)
 		case "/v1/organization/org123/clip/clip-123":
 			fmt.Fprint(w, `{
--- a/origin/origin_test.go
+++ b/origin/origin_test.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"net/http/httptest"
 	"net/url"
 	"testing"
 	"time"
@@ -160,6 +161,9 @@ func TestOrigin_Configure(t *testing.T) {
 		t.Error("Unable to make test urls")
 	}
 
+	cfg, teardown := configMockPropellerAPI()
+	defer teardown()
+
 	tests := []struct {
 		name      string
 		path      string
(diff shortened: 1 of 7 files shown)


origin/origin_test.go
```diff
@@ -217,3 +249,58 @@ func TestOrigin_Configure(t *testing.T) {
 
 	}
 }
+
+// configMockPropellerAPI returns a config.Config object with a mocked version of Propeller
+// API that returns hard-coded responses
+//
+// Make sure to call teardown() when the test is over
+func configMockPropellerAPI() (cfg config.Config, teardown func()) {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
zsiec, line 258: 👍 on using interfaces to avoid the need for a construction like this

origin/origin_test.go
```diff
@@ -217,3 +249,58 @@ func TestOrigin_Configure(t *testing.T) {
...
+				"status": "created",
+				"url": "http://cdn.com/clip-123.m3u8"
+			}`)
+		default:
+			w.WriteHeader(404)
+			fmt.Fprintf(w, "unknown url to mock propeller api: %v", r.URL)
+		}
+	}))
+
+	tsURL, _ := url.Parse(ts.URL)
```
zsiec, line 301: even in tests, I think it's prudent to check errors and fail the test

origin/propeller.go
```diff
@@ -18,16 +19,16 @@ type Propeller struct {
...
-	// Propeller channels can be requested in multiple formats.
-	// When split, the parts in the path should evaluate to the following:
-	// /propeller/orgID/channelID.m3u8
-	// /propeller/orgID/clip/clipID.m3u8
-	// /propeller/orgID/channelID/outputID.m3u8
-	parts := strings.Split(path, "/")
-
-	if len(parts) < 4 {
-		err := fmt.Errorf("propeller origin: request format is not `/propeller/orgID/channelID.m3u8`")
+	urlValues, err := parsePropellerPath(path)
```
zsiec, line 30: Unrelated directly to these changes, it may make sense to eventually be more explicit here perhaps adopting functionality similar to http.ServeMux if we expect this to grow at all in the future.

origin/propeller.go
```diff
@@ -36,13 +37,14 @@ func configurePropeller(c config.Config, path string) (Origin, error) {
 		return &Propeller{}, err
 	}
 
-	orgID := parts[2]
+	orgID := urlValues["orgID"]
+	channelID := urlValues["channelID"]
+	clipID := urlValues["clipID"]
```
zsiec, line 42: one suggestion would be to hoist these values into constants used here and in the regex capture group definition to avoid them diverging accidentally

origin/propeller.go
```diff
@@ -76,6 +78,27 @@ func NewPropeller(c config.Config, orgID string, endpointID string, get fetchURL
...
+			continue
+		}
+		for i, name := range pattern.SubexpNames() {
+			if i != 0 {
+				values[name] = match[i]
+			}
+		}
+		return values, nil
+	}
+	return map[string]string{}, errors.New("propeller origin: request format is not `/propeller/orgID/channelID.m3u8`")
```
zsiec, line 99: Another benefit of being more explicit: this error message seems to mention one specific format when more than one is really supported and that could cause confusion to a consumer.

Also, is there any reason to create and return a map here on error? Why not just `nil`?

origin/origin_test.go
```diff
@@ -217,3 +249,58 @@ func TestOrigin_Configure(t *testing.T) {
 
 	}
 }
+
+// configMockPropellerAPI returns a config.Config object with a mocked version of Propeller
+// API that returns hard-coded responses
+//
+// Make sure to call teardown() when the test is over
+func configMockPropellerAPI() (cfg config.Config, teardown func()) {
```
krystalmejia24, line 257: I do really love these sets of tests you had written here. Initially, my idea was eventually to mock the client itself as you said had stated in your notes 

> If Bakery works with a propeller.Client interface then it's easy to swap
with a mock implementation for tests.

If I had more time (lazy excuse), this was the direction I was planning on going. When I first created the origin package, propeller-go client, I didn't know enough at the time to do that. I did encourage this approach for goofy so @sam-eng  can write her tests much easier. If you want to create the interface and refactor the tests, I'd be all for it! But this is a great way to ensure the end to end functionality when configuring an origin channel.

At some point I considered putting this mocked client under a `test/` package in the propeller-go client: https://github.com/cbsinteractive/goofy/blob/schedule-api/schedule/server/handlers_test.go#L17
so that all of our projects that interface with the client can just import it but im not sure if i was being overzealous with that approach.

igorsobreira:
Summary of changes:

**Refactor URL dispatcher for Propeller origin**

The propeller origin accepts multiple url formats, and with a new format being added I felt that the [current approach of `strings.Split(path, "/")`](https://github.com/cbsinteractive/bakery/blob/d7cdc02492e64204d5a9f46537962ea94901d8ad/origin/propeller.go#L27) was getting hader to read.

So I added a [list of patterns we accept using regex](https://github.com/cbsinteractive/bakery/blob/0b5ae30442065c737409ceb57af7016da053e869/origin/propeller.go#L15-L19) and a [function `parsePropellerPath`](https://github.com/cbsinteractive/bakery/blob/0b5ae30442065c737409ceb57af7016da053e869/origin/propeller.go#L90-L109) to perform the match and extract relevant fields from the URL.

It's still very simple. For instance, we only know [which URL matched based on the fields returned](https://github.com/cbsinteractive/bakery/blob/0b5ae30442065c737409ceb57af7016da053e869/origin/propeller.go#L43-L55). For now this looked enough but a second step would be to give a "name" for each URL and return it so the caller knows which one, or dispatch directly to a method associated to that pattern. Doesn't feel like we need this complexity yet.

**Refactor `fetchURL` functions. Add new one for Channel Output**

`origin/propeller.go` [had a `fetchURL` function type](https://github.com/cbsinteractive/bakery/blob/d7cdc02492e64204d5a9f46537962ea94901d8ad/origin/propeller.go#L19) to allow both implementations [`getPropellerChannelURL`](https://github.com/cbsinteractive/bakery/blob/d7cdc02492e64204d5a9f46537962ea94901d8ad/origin/propeller.go#L79-L110) and [`getPropellerClipURL`](https://github.com/cbsinteractive/bakery/blob/master/origin/propeller.go#L112-L134) to be used by [`NewPropeller`](https://github.com/cbsinteractive/bakery/blob/d7cdc02492e64204d5a9f46537962ea94901d8ad/origin/propeller.go#L59-L62).

I changed that to [an interface `urlGetter`](https://github.com/cbsinteractive/bakery/blob/0b5ae30442065c737409ceb57af7016da053e869/origin/propeller.go#L119-L121). Now we have [`channelURLGetter`](https://github.com/cbsinteractive/bakery/blob/0b5ae30442065c737409ceb57af7016da053e869/origin/propeller.go#L123-L154) and [`clipURLGetter`](https://github.com/cbsinteractive/bakery/blob/0b5ae30442065c737409ceb57af7016da053e869/origin/propeller.go#L207-L233). And, the purpose of the PR, a new [`outputURLGetter`](https://github.com/cbsinteractive/bakery/blob/0b5ae30442065c737409ceb57af7016da053e869/origin/propeller.go#L156-L189)

**origin.Configure() functional tests**

You guys already saw the tests I added using a mocked Propeller API server. After working on `origin/propeller.go` I noticed there were unit tests for it. So I updated the `origin/origin_test.go` tests to [go over only the happy path for Channel, Clip, and Output cases](https://github.com/cbsinteractive/bakery/blob/0b5ae30442065c737409ceb57af7016da053e869/origin/origin_test.go#L174-L191). 

For those tests we mock the Propeller API itself and call `origin.Configure()`. Those are important to have at least once for each propeller entity because they check if we are handling the different URLs correctly

origin/propeller.go
igorsobreira: Yes, that would be the next step. Right now it's easy to identify which one matched based on the fields provided so I didn't want to add unnecessary complexity

origin/origin_test.go
igorsobreira: I didn't check because an error there would mean Go's `httptest.NewServer()` returned an invalid url, which is very unlikely.
5bc757d3957ee403b46afa8e4454147417f13270

origin/propeller.go
igorsobreira: I'm not sure I understand the suggestion. What do you mean by "hoist these values into constants [...]"?

origin/propeller.go
igorsobreira: Good catch, I updated the error message: f761310b916301a9ca40911735d6be75c91d93e2

I used to always return `nil` when an error is being returned, but I noticed [blank values being returned with errors](https://github.com/cbsinteractive/bakery/blob/d7cdc02492e64204d5a9f46537962ea94901d8ad/origin/propeller.go#L71) instead of `nil` so I tried to follow the standard. But since you asked then it's probably not a standard, so I changed to `nil` :)

origin/origin_test.go
igorsobreira: I like to have some of those higher level tests to make sure all objects are being used correctly then test all the small details with unit tests.

My bad for not seeing the unit tests earlier on `origin/propeller_test.go`.

So the approach I take for these unit tests is I specify an interface with a subset of methods used by the propeller client:
https://github.com/cbsinteractive/bakery/blob/0b5ae30442065c737409ceb57af7016da053e869/origin/propeller.go#L111-L115

and I use that inside my module
https://github.com/cbsinteractive/bakery/blob/0b5ae30442065c737409ceb57af7016da053e869/origin/propeller.go#L120
this makes it easy to replace for a mock client.

But I think a better approach would be for the client itself to give us an interface, either `propeller.Client` is an interface (and the client is a `DefaultClient` type) or, if the `propeller-go` only exposes a client the interface could be `propeller.Interface`. This way Bakery's mock can implement that interface and doesn't need the `propellerClient` I created.

If `propeller-go` gives us an interface then it could also provide a Mock implementation of that client like you suggested.

igorsobreira:
Question: how to solve this problem on the CI?
```
Run go test --race -v -coverprofile=coverage.out ./...
verifying github.com/cbsinteractive/propeller-go@v0.0.0-20200503222720-e53e98ec6b80/go.mod: github.com/cbsinteractive/propeller-go@v0.0.0-20200503222720-e53e98ec6b80/go.mod: reading https://sum.golang.org/lookup/github.com/cbsinteractive/propeller-go@v0.0.0-20200503222720-e53e98ec6b80: 410 Gone
```

I had the same problem locally and setting `GOPRIVATE` solved it
```
$ go env |grep -i private
GOPRIVATE="github.com/cbsinteractive/propeller-go"
```

origin/origin_test.go
igorsobreira: Which interface are you referring to?

zsiec:
> Question: how to solve this problem on the CI?
> 
> ```
> Run go test --race -v -coverprofile=coverage.out ./...
> verifying github.com/cbsinteractive/propeller-go@v0.0.0-20200503222720-e53e98ec6b80/go.mod: github.com/cbsinteractive/propeller-go@v0.0.0-20200503222720-e53e98ec6b80/go.mod: reading https://sum.golang.org/lookup/github.com/cbsinteractive/propeller-go@v0.0.0-20200503222720-e53e98ec6b80: 410 Gone
> ```
> 
> I had the same problem locally and setting `GOPRIVATE` solved it
> 
> ```
> $ go env |grep -i private
> GOPRIVATE="github.com/cbsinteractive/propeller-go"
> ```

I think we've been using `go env -w "GONOSUMDB=github.com/cbsinteractive"` in CI for this kind of thing. https://github.com/cbsinteractive/mediahub/blob/master/Dockerfile#L13

origin/propeller.go
zsiec: `"orgID:`, `"channelID"` and `"clipID"` could be defined as constants used in the regex definitions and in these map lookups

origin/propeller.go
zsiec: Gotcha, yeah makes sense to look at project context! In this case, I think we can advocate for always returning the "zero" value (nil if pointer, slice, map, channel etc, empty struct if not a pointer) here and in https://github.com/cbsinteractive/bakery/blob/d7cdc02492e64204d5a9f46537962ea94901d8ad/origin/propeller.go#L71 and everywhere else :)

origin/origin_test.go
krystalmejia24: Yea my thought process here was heading towards `propeller-go` supplying us an interface so that it could provide those mock implementations since we have at least two code bases using this. 

And no worries on missing those earlier tests. These changes make the code base so much cleaner and easier to read. lets get rid of my elementary code!

origin/origin_test.go
zsiec: I agree, quite unlikely :) I think it sets a good precedent though. In this case, rather than `panic` we could supply a `*testing.T` to this method and simply call `t.Fail` which would be less surprising than a panic during testing.

origin/origin_test.go
zsiec: I don't think the interface exists yet, I was just agreeing that a client interface (once it exists) will help simplify here :). We do this with [most](https://github.com/cbsinteractive/drm-svc/blob/master/client/drmsvc/client.go#L18) of our [clients](https://github.com/cbsinteractive/transcode-orchestrator/blob/master/client/transcoding/client.go#L12)

origin/propeller_test.go
```diff
@@ -1,250 +1,378 @@
...
-			} else if err == nil && tc.expectError {
-				t.Errorf("getClipURL() expected error, got nil")
+			u, err := tc.getter.GetURL(tc.client)
+			if err != nil && tc.expectErrStr == "" {
+				t.Errorf("returned unexpected error: %q", err)
+			} else if err == nil && tc.expectErrStr != "" {
+				t.Errorf("expected error, got nil")
+			} else if err != nil && err.Error() != tc.expectErrStr {
+				t.Errorf("got wrong error. expected: %q, got %q", tc.expectErrStr, err.Error())
 			}
```
zsiec, line 219: If we expect an error, do we want to continue with the test here? In other projects we've use a construct like:

```go
if err != nil {
	if g, e := err.Error(), tt.wantErr; g != e {
		t.Errorf("Func() wrong error returned, got: %v, want: %v", g, e)
	}
	return
} else if tt.wantErr != "" {
	t.Error("Func() expected an error, got nil")
	return
}

```

origin/origin_test.go
zsiec: I see you added them later on in this PR 💯

origin/origin_test.go
igorsobreira: 8b9cf2b

origin/propeller.go
igorsobreira: Yep, agree

origin/propeller_test.go
igorsobreira: No strong personal preferences here. In this case, I kept going because the returned URL can give a hint on why the error did not match. And I know the return value will not be nil (it's an empty string)

But as a general rule, I agree, it's dangerous to try and use the value is the error doesn't match

zsiec: :+1: We should probably update `t.Errorf("expected error, got nil")` to use `t.Error` if we're not passing args to format.

igorsobreira: Yep, good catch 96fd547

PR #103, opened 2020-05-06
Trim filters with carrier captions playlist

keithrichardscbs:


keithrichardscbs:
@krystalmejia24 So the failed codecov is pretty obnoxious. It's missing error cases that are difficult to produce. I can address it if you like by monkey patching the calls and checking for expected errors, I'm just not sure how much that is worth the time.

filters/hls.go
```diff
@@ -434,3 +440,31 @@ func getAbsoluteURL(path string) (*url.URL, error) {
...
+		if alreadyProcessedAlternative(alt.GroupId, trimmedAlternatives) {
+			continue
+		}
+		if alt.Type == "SUBTITLES" {
+			auri, err := h.normalizeTrimmedVariant(filters, alt.URI)
+			if err != nil {
+				return trimmedAlternatives, err
+			}
+			alt.URI = auri
+			trimmedAlternatives = append(trimmedAlternatives, alt.GroupId)
```
krystalmejia24, line 466: what if we hold `trimmedAlternatives` in a map? if we use something like this, `map[string]struct{}`
then its a quick lookup as opposed to iterating through a list.

`
if _, found := trimmedAlternatives[v.Subtitles]; found {
		continue
}
`

filters/hls_test.go
```diff
@@ -1524,6 +1563,15 @@ http://localhost:8080/t(10000,100000)/tags(ads)/aHR0cHM6Ly9leGlzdGluZy5iYXNlL3Bh
 			expectManifestContent: masterManifestWithAbsoluteURLs,
 			config:                config.Config{Hostname: "bakery.cbsi.video"},
 		},
+		{
+			name:                  "when media tag is present, filter and replace media url with base64 encoded string",
```
krystalmejia24, line 1567: Do we need to test for multiple subtitle groups? I know Im thinking long term for things like international, but should be easy enough to add an extra subs group and attach them to differing variants. It will test the `alreadyProcessed` functionality a lot better too.

keithrichardscbs: Yeah I can do that. The `alreadyProcessed` stuff was there because each variant had the same alternative and I had to prevent them all from reprocessing. But multiple alternatives needs to be tested too.

filters/hls.go
```diff
@@ -62,6 +62,8 @@ func (h *HLSFilter) FilterManifest(filters *parsers.MediaFilters) (string, error
 	filteredManifest := m3u8.NewMasterPlaylist()
 	filteredManifest.Twitch = manifest.Twitch
 
+	trimmedAlternatives := make(map[string]bool)
```
krystalmejia24, line 65: do we need to initialize this here? Wondering if we can just have the have the scope of this live within the condition checking for the trim filter itself.

PR #105, opened 2020-05-12
Logging

krystalmejia24:
Updated logging lib to zerolog from logrus. 

When @zsiec showed me this I was sold. https://github.com/rs/zerolog#benchmarks

It has seamless integration with net/http package so it gives us more visibility into our requests. The json formatter that works well with NewRelic. We should be able to create better dashboards once I have access. 

Also improved some of the ctx logs within the origin package when configuring propeller channels.


origin/propeller.go
```diff
@@ -146,11 +145,14 @@ func (g *channelURLGetter) getURL(channel propeller.Channel) (string, error) {
 	if channel.Captions {
 		return channel.CaptionsURL, nil
 	}
-	playbackURL, err := channel.URL()
-	if err != nil {
-		return "", fmt.Errorf("parsing channel url: %w", err)
+	if channel.PlaybackURL == "" {
+		if channel.Outputs != nil {
+			return "", fmt.Errorf("channel has multiple outputs. Expect request format /propeller/org-id/channel-id/output-id.m3u8")
+		}
```
igorsobreira, line 151: @krystalmejia24 I think using `len(channel.Outputs) > 0` might be better, it would also handle the case where `Outputs` array is set but empty

https://play.golang.org/p/PqCqLRJhLiM

origin/propeller.go
```diff
@@ -199,7 +201,11 @@ func handleGetUrlChannelNotFound(err error, orgID string, channelID string, clie
 			orgID:  orgID,
 			clipID: fmt.Sprintf("%v-archive", channelID),
 		}
-		return clipGetter.GetURL(client)
+		archive, clipErr := clipGetter.GetURL(client)
+		if clipErr != nil {
+			return "", fmt.Errorf("Channel %v Not Found", channelID)
```
igorsobreira, line 206: Have you considered using `%w` to avoid silencing the `clipErr` completely? `clipGetter.GetURL()` can fail due to other reasons besides channel not found

Maybe something like this would allow the information from `clipErr` to not be lost:
```go
return "", fmt.Errorf("Channel %v Not Found: %w", channelID, clipErr)
```

origin/propeller.go
```diff
@@ -5,17 +5,15 @@ import (
...
 	"github.com/cbsinteractive/bakery/config"
 	propeller "github.com/cbsinteractive/propeller-go/client"
 )
 
 // propellerPaths defines the multiple path formats allowed for propeller entities in Bakery
 var propellerPaths = []*regexp.Regexp{
 	regexp.MustCompile(`/propeller/(?P<orgID>.+)/clip/(?P<clipID>.+).m3u8`),
-	regexp.MustCompile(`/propeller/(?P<orgID>.+)/(?P<channelID>.+)/(?P<outputID>.+).m3u8`),
-	regexp.MustCompile(`/propeller/(?P<orgID>.+)/(?P<channelID>.+).m3u8`),
+	regexp.MustCompile(`/propeller/(?P<orgID>.+)/(?P<channelID>.+)/(?P<outputID>.+).(m3u8|mpd)`),
```
igorsobreira, line 15: I noticed that the extension `m3u8` or `mpd` are not being provided to `outputURLGetter`. This won't cause problems but it will have a funny side effect: if that specific output is a HLS stream, then both URLs `.m3u8` and `.mpd` will work in bakery but return the HLS manifest

If we name this regex group and match the value against [`Channel.Outputs[n].Type`](https://github.com/cbsinteractive/propeller-go/blob/f613608c02512540783721e02a7994bba5f935a2/client/channel.go#L13
) (which is either `"hls"` or `"dash"`) then we can validate and return an error if the format you're asking does not match.


[Possible `Type` values for output](https://github.com/cbsinteractive/propeller/blob/122fd8016e54de4458fa856faa0e0d73cfba3dbe/app/api/models.py#L650-L675): `"rtp"`, `"hls"` and `"dash"`

origin/propeller.go
```diff
@@ -5,17 +5,15 @@ import (
...
+	regexp.MustCompile(`/propeller/(?P<orgID>.+)/(?P<channelID>.+).(m3u8|mpd)`),
```
igorsobreira, line 16: At this moment in Propeller, the only way to create a channel with DASH is to use `"outputs"`, in that case, you will be using the output id URL.

igorsobreira:
@krystalmejia24 sorry, I know it already merged but I added some comments just to clarify the "outputs" behavior in Propeller.

(related https://github.com/cbsinteractive/propeller/pull/1015)

origin/propeller.go
krystalmejia24: Thank you! did not consider a set but empty slice

origin/propeller.go
krystalmejia24: I actually intentionally did this since this was specifically for the automated archive request. I didnt want to respond with clip error when the request was for a channel. But to your point, Im now silencing a possible archive error!

I believe i can grab archive true from channel directly, so if we add that condition so we only call `handleGetUrlChannelNotFound` and maybe rename to `getChannelArchive`, I can update the message to something like: `return "", fmt.Errorf("Channel %v Archive Error: %w", channelID, clipErr)`

krystalmejia24:
@igorsobreira appreciate you checking even after i merged! I needed to merge this for the logging work. In the future I'll know better than to stack 2 issues into one PR. 

I looked through some of the comments to make sure nothing would be breaking so I'll have to go back in and update the DASH functionality that does not work right now.

origin/propeller.go
igorsobreira: Gotcha! Makes sense, you're handling a channel request so you're reporting a "channel not found" error, I think it's fine the way it is then.

I like the idea of checking `archive: true` flag from the channel, it can also improve performance avoiding a request if it's false. It would be a nice improvement at some point.

Issue #107, opened 2020-05-12
Trim filter should include all content requested.


keithrichardscbs:
Suppose a manifest looks like this:

> #EXT-X-VERSION:3                                                                                                                                                                                   
> #EXT-X-TARGETDURATION:7
> #EXT-X-PROGRAM-DATE-TIME:2020-05-12T15:30:14.000000+00:00
> #EXTINF:6.006,                                                                                                                                                                                     
> en/cbssc0a7_1_20200512T153015_112660.vtt
> #EXT-X-PROGRAM-DATE-TIME:2020-05-12T15:30:20.006000+00:00                                                                                                                                          
> #EXTINF:6.006,                                                                                                                                                                                     
> en/cbssc0a7_1_20200512T153021_112661.vtt                                                                                                                                                           

If a filter request comes in for 15:30:15 -- 15:30:26? Right now the behavior is to only include `en/cbssc0a7_1_20200512T153021_112661.vtt`. However, this means the request for 11 seconds of content only returns 6 seconds. Furthermore, the contents start 5 seconds later than requested. All of this when the manifest does in fact fully index the time frame specified.

I propose in the example above the request for 15:30:15 -- 15:30:26 should return the entire manifest (include any segments that cover any of the specified range).

PR #108, opened 2020-05-13
Include all segments that have ANY content in the trim filter range (…

keithrichardscbs:
…#107)


docs/filters/trim.md
```diff
@@ -5,7 +5,7 @@ nav_order: 4
 ---
 
 # Trim
-An **INCLUSIVE RANGE** of segments to **INCLUDE** in the modified variant playlist. Segments outside this range will be filtered out. The Playlist returned will be a Video on Demand Playlist. 
+An **INCLUSIVE RANGE** of segments to **INCLUDE** in the modified variant playlist. Segments that contain ANY amount of content included in the range will be included. Segments with NO content in the range will be filtered out. The Playlist returned will be a Video on Demand Playlist.
```
krystalmejia24, line 8: THANKS FOR UPDATING THE DOCS! :)

filters/hls.go
```diff
@@ -372,7 +379,15 @@ func (h *HLSFilter) filterRenditionManifest(filters *parsers.MediaFilters, m *m3
...
-		append = inRange(filters.Trim.Start, filters.Trim.End, int(segment.ProgramDateTime.Unix()))
+		segmentTimestamp := int(segment.ProgramDateTime.Unix())
+
+		isInRange := inRange(filters.Trim.Start, filters.Trim.End, segmentTimestamp)
+		if shouldIncludePreviousSegmentInTrimmedRenditionManifest(append, isInRange, i, segmentTimestamp, filters.Trim.Start) {
+			if err := appendSegment(h.manifestURL, m.Segments[i-1], filteredPlaylist); err != nil {
+				return "", fmt.Errorf("trimming segments: %w", err)
+			}
+		}
+		append = isInRange
```
krystalmejia24, line 390: one thing to keep in mind is that the segments are held in a circular buffer so lets always check if the previous segment is nil. https://github.com/grafov/m3u8/issues/127 

Also, im wondering if we can just update the `append` mode by checking the current segment end? 
```
	segmentTimestamp := int(segment.ProgramDateTime.Unix())
	if append = inRange(filters.Trim.Start, filters.Trim.End, segmentTimestamp); !append {
		currentSegmentEnd := segmentTimestamp + int(segment.Duration)
		append = inRange(filters.Trim.Start, filters.Trim.End, currentSegmentEnd)
	}
```

less dependencies on index, nil segments.

Not sure if yout think we need to introduce `inclusiveInRange` and `exclusiveInRange` im assuming the current `InRange` function, which is inclusive, is enough for checking the end time of the segment.

filters/hls_test.go
```diff
@@ -1669,6 +1669,31 @@ https://existing.base/path/chan_1/chan_1_20200311T202813_1_00024.ts
 #EXTINF:6.000,
 https://existing.base/path/chan_1/chan_1_20200311T202818_1_00025.ts
 #EXT-X-ENDLIST
+`
```
krystalmejia24, line 1672: lets also add tests here https://github.com/cbsinteractive/bakery/blob/6137a3f8a9bac78a29b83bf43a61610a67f6478c/filters/hls_test.go#L1784

when adding tests cases to this it would be good to really exercise those boundaries. Something like having the ad tags, which have no PDT, sit right after a segment whose exclusively in range.

filters/hls.go
keithrichardscbs: Yeah I like this. However, we will have to keep the check against the start time because if the segment ends right at the start of the range, inRange would return true (and we want to exlude it in that case).

keithrichardscbs:
@krystalmejia24 I think you may have force pushed over the changes you requested. Suddenly the commits don't match my local branch.

krystalmejia24:
@keithrichardscbs i totally did do that! I accidentally rebased on the wrong commit as I had pulled your changes in previously, but never pulled in the newest commit with change we discussed. 

I have cleaned that up https://github.com/cbsinteractive/bakery/pull/109/files

as you can see its just your two changes and the rebase i made against master branch. Im pulling your changes into that branch as they are all related. Once i can evaluate both working with eachother will merge to dev!

PR #112, opened 2020-06-05
Redundant manifests

krystalmejia24:


parsers/urlparser.go
```diff
@@ -185,6 +186,8 @@ func URLParse(urlpath string) (string, *MediaFilters, error) {
 				fr := strings.ReplaceAll(framerate, ":", "/")
 				mf.FrameRate = append(mf.FrameRate, fr)
 			}
+		case "sp":
+			mf.SP = true
```
krystalmejia24, line 190: @flavioribeiro not sure how i feel about this naming. `SP` is a temp for single pipeline but i wasn't sure if thats where we wanted to go.

filters/hls.go
```diff
@@ -448,20 +493,11 @@ func getAbsoluteURL(path string) (*url.URL, error) {
 	return url.Parse(absoluteURL)
 }
 
-func alreadyProcessedAlternative(groupId string, processedGroupIds []string) bool {
```
krystalmejia24, line 451: i did a quick refactor of this though it does not pertain to the work. maps are better for this kind of lookup rather than iterating through slices

parsers/urlparser.go
```diff
@@ -185,6 +186,17 @@ func URLParse(urlpath string) (string, *MediaFilters, error) {
 				fr := strings.ReplaceAll(framerate, ":", "/")
 				mf.FrameRate = append(mf.FrameRate, fr)
 			}
+		case "dw":
```
igorsobreira, line 189: Is the current approach with other filters to use short abbreviations like this one? I think `dewave` would be more readable. Just curious is there is already a standard way.

origin/origin.go
```diff
@@ -15,14 +15,21 @@ import (
...
-	Origin string
-	URL    url.URL
+	Host string
+	URL  url.URL
+}
+
+//ManifestInfo holds http response info from manifest request
+type ManifestInfo struct {
+	Manifest     string
+	LastModified string
```
igorsobreira, line 31: Have you considered having this field already as a `time.Time` object? I noticed the only use for it is parsing the string value.

filters/hls.go
```diff
@@ -470,8 +506,70 @@ func (h *HLSFilter) normalizeTrimmedVariantAlternatives(filters *parsers.MediaFi
...
+
+	playlist := v.(*m3u8.MediaPlaylist)
+
+	lastModified, err := http.ParseTime(lastModifiedHeader)
+	if err != nil {
+		return false, err
+	}
+
+	segDurationX2 := time.Second * time.Duration(playlist.TargetDuration*2)
+	diff := time.Now().Sub(lastModified)
```
igorsobreira, line 572: have you taken into consideration in tests a `lastModified` as UTC while `time.Now()` is local time? Does go ensure the difference is the same regardless of the timezone or is it something we need to handle ourselves?

parsers/urlparser.go
krystalmejia24: the approach is short abbreviations. these are used directly in the url.

filters/hls.go
krystalmejia24: @igorsobreira I was doing that! 

Learned from @as that calling `UTC()` just changes the location to nil which wouldn't affect `Sub()` call since GO uses unix timestamps under the hood.

you can see the logs i was dumping when running my comparison. One of the diffs is w/ UTC. 

```
lastModified:	2020-06-07 16:00:03 +0000 UTC
timenow:	2020-06-07 12:00:07.383584 -0400 EDT m=+14.352724774
diff1:	4.383584s
diff2:	4.383584s
segDurationx2:	16s
```

filters/hls.go
```diff
@@ -31,6 +33,11 @@ var matchFunctions = map[ContentType]func(string) bool{
 	captionContentType: isCaptionCodec,
 }
 
+const (
+	primaryPipeline = "primary"
+	backupPipeline  = "backup"
+)
```
igorsobreira, line 39: Have you considered using a custom type these? Could be type string if you need those string values somewhere, otherwise [`iota` would be enough](https://golang.org/ref/spec#Iota).

It would make `filterPipeline()` API more clear, like
```go
type pipelineType string
const (
    primaryPipeline pipelineType = "master"
    backupPipeline               = "backup"
)
func (h *HLSFilter) filterPipeline(uri string) (pipelineType, error) {
```

igorsobreira:
@krystalmejia24 thanks a lot for working on this so quicky! 👍

igorsobreira:
@krystalmejia24 question about bakery behavior in general: how often we will fetch the master manifest to get a chance to switch back from backup to primary and vice-versa?

filters/hls.go
krystalmejia24: @igorsobreira I keep forgetting about iota and it would actually be good enough. I prefer iota bc of its simplicity but i see the value in adding the custom type if it helps with clairity and readability? how unlcear was this for you? 

```
const (
    primaryPipeline = iota //primary
    backupPipeline  = iota //backup

)
```

origin/origin.go
krystalmejia24: nope hadn't considered. Will make this change since everyone is oging to want to interface with the `time.Time` object not the string.

krystalmejia24:
@igorsobreira good q and i think will default to what @flavioribeiro has to say. While I understand how the player switches between the ABR ladder for the different renditions, my assumption has always been one fetch for master during playback.

filters/hls.go
igorsobreira: Nice. I just meant the `filterPipeline` was returning a `string`, when in fact it returns two very specific string values. I could understand the code just fine, but if you had a more restrictive type in the function signature it would be more clear what will be returned.

flavioribeiro:
that’s correct, one fetch per session

PR #113, opened 2020-06-11
Redundant manifests

krystalmejia24:


filters/hls.go
```diff
@@ -549,6 +549,10 @@ func healthCheckVariant(variantURL string, client config.Client) (bool, error) {
 		return false, fmt.Errorf("checking variant: returning http status of %v", sc)
 	}
 
+	if manifestInfo.LastModified.IsZero() {
```
krystalmejia24, line 552: Im not entirely sure what behavior we want if LastModified time is not set on the variant. do we want to error out? I can add the test case for this if thats behavior desired. I didnt think we'd want to just fallback to other pipeline.

zsiec: I don't think we should fail here, there may be streams that don't provide this info. My gut says we should check both primary and any backups and if none of them have this value, just default to the primary?

PR #114, opened 2020-06-12
logging: consolidate logs

zsiec:
Collapses all logging to a single line per request, enrich the logging context with data as it becomes available.

### DIFF
--- a/handlers/handler.go
+++ b/handlers/handler.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/cbsinteractive/bakery/config"
 	"github.com/cbsinteractive/bakery/filters"
+	"github.com/cbsinteractive/bakery/logging"
 	"github.com/cbsinteractive/bakery/origin"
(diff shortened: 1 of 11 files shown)


handlers/handler.go
```diff
@@ -18,36 +19,30 @@ func LoadHandler(c config.Config) http.Handler {
 		c.Client.SetContext(r)
 
 		w.Header().Set("Access-Control-Allow-Origin", "*")
-		ctxLogger := c.Logger.With().Str("uri", r.RequestURI).Logger()
-		ctxLogger.Info().
```
krystalmejia24, line 22: do we want remove the "received request" log line?

zsiec: I think we do want to remove this, we already log all this at the end of the request in our `AccessHandler` middleware. What value does this line have that the other line can't provide?

krystalmejia24: makes sense!

PR #124, opened 2020-08-28
Cleanup context

krystalmejia24:


origin/origin.go
```diff
@@ -76,17 +76,17 @@ func (d *DefaultOrigin) GetPlaybackURL() string {
...
 	req, err := http.NewRequest(http.MethodGet, manifestURL, nil)
 	if err != nil {
 		return ManifestInfo{}, fmt.Errorf("generating request to fetch manifest: %w", err)
 	}
 
-	ctx, cancel := context.WithTimeout(client.Context, client.Timeout)
+	ctx, cancel := context.WithTimeout(parentCtx, client.Timeout)
 	defer cancel()
 
 	resp, err := client.Do(req.WithContext(ctx))
```
krystalmejia24, line 92: so im looking for the best way to manage this. I'm looking at using `http.NewRequestWithContext` to simplify how we use this and make it easier to debug where the cancelation is coming from, but I'm not entirely sure this could cause problems down the road. 

FYI I also made that change here: https://github.com/cbsinteractive/propeller-go/pull/11/commits/0108b0d3d3b7fd18d42f167bfbc42d71cd911985 thought I have similar q's and doubts.

filters/dash.go
```diff
@@ -37,7 +38,7 @@ func (d *DASHFilter) GetMaxAge() string {
 }
 
 // FilterManifest will be responsible for filtering the manifest according  to the MediaFilters
-func (d *DASHFilter) FilterManifest(filters *parsers.MediaFilters) (string, error) {
+func (d *DASHFilter) FilterManifest(ctx context.Context, filters *parsers.MediaFilters) (string, error) {
```
as, line 41: This function doesn't look like it needs a context because it doesn't check for a cancellation signal and doesn't pass the context to anywhere else. A good rule of thumb is to minimize the number of places we have a context argument and only have it for functions that perform network calls or other long operations.

krystalmejia24: @as the HLS filter will need this bc we need to make network requests for redundant pipelines we pass through. Since `FilterManifest` is part of the interface method I figured no harm in passing this here. 

Any idea in if there is a better way to handle this?

as: Good question, I think in the interest of time we can just go with this signature for now and circle back later. In which case I would rewrite the function signature as `FilterManifest(_ context.Context, filters *parsers.MediaFilters) (string, error)` to indicate that the `ctx` isn't used: makes things easier to reason about.

allisonnicoledeal:
👍 Thanks for the update here!

PR #135, opened 2021-01-31
implement PreventHTTPStatusError filter (fixes #134)

### LINKED ISSUE #134: Bakery to front dictate for Webvtt failures
As CBS Sports, I can use bakery to front all dictate URLS so that I can protect against content failure stemming from WebVTT failures.

**Acceptance Criteria:**

- Bakery URLs are used to proxy Dictate URLs
- In the event dictate fails or WebVTT files are not present, Bakery returns empty manifests
- Change is validated on CBSi Avia Player

**Requirements:**

<span data-href="#aha_task_list_start" data-version="v2" data-aha="Do not remove this tag. Used for Aha! syncing."/>

- [x] <span data-href="#aha_task_list_BAKE-33-1" data-version="v2" data-aha="Do not remove this tag. Used for Aha! syncing."/> implement PreventHTTPStatusError filter
    <br/>
- [x] Deploy to production
<span data-href="#aha_task_list_end" data-version="v2" data-aha="Do not remove this tag. Used for Aha! syncing."/>

Aha! Link: https://viacomcbs-digital.aha.io/features/BAKE-33

flavioribeiro:
fix #134


handlers/handler.go
```diff
@@ -44,6 +44,17 @@ func LoadHandler(c config.Config) http.Handler {
 
 		//throw status error if not 2xx
 		if manifestInfo.Status/100 > 3 {
+			if mediaFilters.PreventHTTPStatusError {
+				switch mediaFilters.Protocol {
+				case parsers.ProtocolHLS:
+					w.Header().Set("Content-Type", "application/x-mpegURL")
+					fmt.Fprint(w, filters.EmptyHLSManifestContent)
+				case parsers.ProtocolVTT:
```
zsiec, line 52: The switch now on line 67 will not set a filter for the `vtt` "protocol" causing a panic when `f.FilterManifest` is called on line 77.

zsiec: Also, I wonder if we should rename `manifestInfo` as it can contain media/variant manifest content, mpd content or VTT cues?

docs/filters/prevent-http-status-error.md
```diff
@@ -0,0 +1,50 @@
+---
+title: Prevent HTTP Error
```
zsiec, line 2: <3 docs!!

filters/hls.go
```diff
@@ -18,6 +18,8 @@ import (
 
 type execPluginHLS func(variant *m3u8.Variant)
 
+const EmptyHLSManifestContent = "#EXTM3U"
```
zsiec, line 21: Just curious, I wonder if there's any case we've seen where an "empty" HLS would avoid an error in playback.

origin/origin.go
```diff
@@ -39,14 +40,13 @@ func Configure(ctx context.Context, c config.Config, path string) (Origin, error
...
-		variantURL, err := decodeVariantURL(parts[1])
+	// check if path is base64 encoded url
+	if strings.Count(path, "/") == 1 && (strings.HasSuffix(path, ".m3u8") || strings.HasSuffix(path, ".vtt")) {
+		decodedPath, err := trimAndDecodePath(strings.TrimPrefix(path, "/"))
 		if err != nil {
-			return &DefaultOrigin{}, fmt.Errorf("decoding variant manifest url %q: %w", path, err)
+			return &DefaultOrigin{}, fmt.Errorf("decoding base64 url %q: %w", path, err)
 		}
-		path = variantURL
+		path = decodedPath
```
zsiec, line 49: It's odd we call this `path`, in many (if not most?) cases, won't this be the full url?

origin/origin.go
```diff
@@ -115,9 +115,9 @@ func fetch(ctx context.Context, client config.Client, manifestURL string) (Manif
 	}, nil
 }
 
-func decodeVariantURL(variant string) (string, error) {
-	variant = strings.TrimSuffix(variant, ".m3u8")
-	url, err := base64.RawURLEncoding.DecodeString(variant)
+func trimAndDecodePath(encodedPath string) (string, error) {
+	encodedPath = strings.TrimSuffix(encodedPath, filepath.Ext(encodedPath))
```
zsiec, line 119: we might want `path.Ext` as `filepath` assumes path separators based on the os execution context.

zsiec (changes requested):
Added a few comments, I assume re-writing the vtt subtitle manifests to use bakery urls for vtt segments is out-of-scope, perhaps Dictate will already do this ahead of time?

filters/vtt.go
```diff
@@ -0,0 +1,3 @@
+package filters
+
+const EmptyVTTContent = "WEBVTT"
```
keithrichardscbs, line 3: In Dictate this is what we return for empty vtt:
`"WEBVTT\nX-TIMESTAMP-MAP=LOCAL:00:00:00.000,MPEGTS:0\n\nNOTE No timecoded transcript received. No captions generated.\n"`

The timestamp stuff I have found helps Apple players. The Note of course we could change to something that would help identify that it came from bakery.

filters/hls.go
flavioribeiro: Anecdotally, a service (like dictate) could error out on creating the `en.m3u8` for the vtt renditions but still retry and create it over time. If we serve an empty manifest instead of erroring the request out, we could potentially still recover that player that initiated the session when the real manifest wasn't still there. If we serve the 404 all the way back, some players will remove or quarantine that rendition during the playback.

Note that a lot of what I wrote isn't tested and based on assumptions. My idea was to defer to the "client" (in this case, dictate) in whether to protect a request or not. Does that make sense?

origin/origin.go
flavioribeiro: good catch

filters/vtt.go
flavioribeiro: pretty useful stuff, thanks @keithrichardscbs

handlers/handler.go
flavioribeiro: > The switch now on line 67 will not set a filter for the vtt "protocol" causing a panic when f.FilterManifest is called on line 77 because f is nil.

The code will never reach line 77 as it will return _something_ if the request to the origin errors out. However, I agree with you things can become hard to manage as we have "filters" that don't act as filters. Should we have a `NewVTTFilter` on `vtt.go` that just returns the plain vtt cue and conform with the others?

> Also, I wonder if we should rename manifestInfo as it can contain media/variant manifest content, mpd content or VTT cues?

I think so. Can we go with `Content`? `Payload`? We'll need to change the `FetchManifest` function on `origin` and a few other things to conform to this change.

flavioribeiro:
Thanks to the comments, @zsiec, and @keithrichardscbs. I updated the PR with the suggestions and replied to some others.

handlers/handler.go
zsiec: Won't it reach line 77 if the origin doesn't error?

zsiec: `OriginResponse` as a name perhaps?

This is the struct as it stands:
```go
//ManifestInfo holds http response info from manifest request
type ManifestInfo struct {
	Manifest     string
	LastModified time.Time
	Status       int
}
```

+1 on renaming the surrounding subroutines/structs

zsiec: >  Should we have a NewVTTFilter on vtt.go that just returns the plain vtt cue and conform with the others?

yeah I think this makes sense, it might be just a passthrough for now, but we can add manipulation there if we ever need it

origin/origin.go
```diff
@@ -42,11 +42,11 @@ func Configure(ctx context.Context, c config.Config, path string) (Origin, error
 
 	// check if path is base64 encoded url
 	if strings.Count(path, "/") == 1 && (strings.HasSuffix(path, ".m3u8") || strings.HasSuffix(path, ".vtt")) {
-		decodedPath, err := trimAndDecodePath(strings.TrimPrefix(path, "/"))
+		url, err := trimAndDecodePath(strings.TrimPrefix(path, "/"))
```
zsiec, line 45: Odd Go thing, I try to avoid using variables that collide with stdlib go package names, this could be `u` or `originURL` or something?

zsiec (changes requested):
Most of my remaining comments are stylistic. Would approve except, I want to confirm the logic is correct in the happy path and won't cause panics.

origin/origin.go
flavioribeiro: totally, thanks for catching this one!

handlers/handler.go
flavioribeiro: this should be solved now

origin/origin.go
zsiec: did we lose changes here? I thought you changed this in recent commit

zsiec (commented):
Added a comment, not sure if we lost some changes in the rebase

origin/origin.go
flavioribeiro: yeah that seems to be the case, I'm seeing the changes in 7f18ab87

flavioribeiro: I cherry picked that one, but very interesting observation, I'm going over everything now to see if we lost changes. The fact is that I coded this one in two different computers, so that might have caused some issues

zsiec (approved):
LGTM!

PR #151, opened 2021-07-02
remove strict requirements around auth and propeller creds

krystalmejia24:
TODO:

- [x] Update README to reflect new env vars and startup process
- [x] update bakery-deploy to enable auth and propeller origin on startup.


config/config.go
```diff
@@ -68,12 +65,12 @@ func (c Config) getLogger() zerolog.Logger {
 
 //ValidateAuthHeader returns key,value or error if not set
 func (c Config) ValidateAuthHeader() error {
-	if c.IsLocalHost() {
+	if c.IsLocalHost() || !c.AuthEnabled {
 		return nil
 	}
 
 	if c.OriginKey == "" || c.OriginToken == "" {
```
vish91, line 72: will origin key every be null ? 
seems like an extra check since it has a default value
```OriginKey string `encovnfig:"ORIGIN_KEY" default:"x-bakery-origin-token"````

**Comment by vish91** [comment]
got much better results with this PR: 👍🏽 
```
docker run --privileged --env BAKERY_ORIGIN_HOST="<something>" -p 8080:8080 -d --name bakery bakery:latest
d3ffdb78f7350aac297a0d2483e1431e67bfa89df103b4c29c8c044a76fa3b61
 
 $ docker ps
CONTAINER ID   IMAGE           COMMAND      CREATED         STATUS         PORTS                                       NAMES
d3ffdb78f735   bakery:latest   "./bakery"   4 seconds ago   Up 3 seconds   0.0.0.0:8080->8080/tcp, :::8080->8080/tcp   bakery

Test:
curl -v localhost:8080/intl_vms/2021/05/27/1902237251849/705381_fp_hls/master.m3u8 -o /dev/null
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8080 (#0)
> GET /intl_vms/2021/05/27/1902237251849/705381_fp_hls/master.m3u8 HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.64.1
> Accept: */*
>
< HTTP/1.1 200 OK
```

File: config/config.go (hunk 1 above)
Comment (krystalmejia24) [reply config/config.go:72]: I think this will get set null if we export the var with a null value. 

Also, we can discuss in more detail later on when it comes to CDN configuration, but we only use auth to ensure all of our requests are served from CDN and we reject any requests directly to the origin. You may want to follow a similar pattern once you deploy your version.

**Review by vish91 — APPROVED** [review]

File: config/config.go [hunk 2]
```diff
@@ -68,12 +65,12 @@ func (c Config) getLogger() zerolog.Logger {
 
 //ValidateAuthHeader returns key,value or error if not set
```
terranvigil, line 66: This comment doesn't match the code - should be 'returns nil'?

krzemienski:
@krystalmejia24 Thanks for this...I was actually working on something locally!

PR #153, opened 2021-07-14
VideoRobot github actions

vigneshrcbs:
- Created an action file for building and deploying the application to VideoRobot google cloud environment.
- Tested using ACT.


.github/workflows/build-vr.yml
```diff
@@ -0,0 +1,79 @@
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
vish91, line 5: on tags only for prod

.github/workflows/build-vr.yml
```diff
@@ -0,0 +1,79 @@
...
+      GCP_PROJECT_NAME: "i-vidbot-dev"
+    steps:
+      - name: checkout
+      # TODO: checkout to master
+        uses: actions/checkout@master
+  
+      - name: Set short-sha
+        id: vars
+        run: echo "::set-output name=sha_short::$(git rev-parse --short HEAD)"
+
```
vish91, line 28: adding a make all step to make sure the app builds fine.

.github/workflows/build-vr.yml
krystalmejia24: We should remove the pull_request condition as it would always deploy to dev once the PR is made. We have a build action in this repo already that should ensure tests are passing before PRs to master are merged. 

I'm unclear as I haven't written an action in like over a year, but if your able to manage multiple push sources like branches AND tags, you can then extend/update the condition on line 49 to check for the tags from the master branch.

If multiple push sources We may need to separate these tasks out into 2 separate actions, something like: `deploy_dev_vr` and `deploy_prod_vr`.

vigneshrcbs: About the pull_request condition, I think it's there to deploy bakery to the Video Robot dev env too after every merge to master. Right now it only deploys to your environment.

vish91: @krystalmejia24  the way we control that is by using github env , so only if that action is approved by someone on the team it will trigger this github action... 
that avoids the concern of creating too many unwanted docker images... people can keep making the PR and anyone from a core team can approve the videorobot github action build when they feel ready. 
There's a github doc explaining this feature.. i couldn't find it but this one explains that too
https://devblogs.microsoft.com/devops/i-need-manual-approvers-for-github-actions-and-i-got-them-now/

vish91: I still have to set that part up in the repo settings along with some GCP secrets... will do that once Vignesh cleans up this pr... 
and yeah you are right if multiple push sources don't work we will ahve to separate it out.. left that for Vignesh to test and see what he will need to do...

vish91 (approved):

krystalmejia24 (approved):

PR #160, opened 2021-08-09
hls: filter bandwidth by average bandwidth first, fallback if average bandwidth is not provided

zsiec:
Fixes: #161


filters/hls_test.go
```diff
@@ -1164,6 +1164,49 @@ http://existing.base/uri/link_13.m3u8
...
+#EXT-X-STREAM-INF:PROGRAM-ID=0,BANDWIDTH=1000,AVERAGE-BANDWIDTH=400,CODECS="avc1.77.30"
+http://existing.base/uri/link_1.m3u8
+#EXT-X-STREAM-INF:PROGRAM-ID=0,BANDWIDTH=2000,AVERAGE-BANDWIDTH=900,CODECS="avc1.77.30"
+http://existing.base/uri/link_2.m3u8`,
+			expectManifestContent: `#EXTM3U
+#EXT-X-VERSION:3
+#EXT-X-STREAM-INF:PROGRAM-ID=0,BANDWIDTH=2000,AVERAGE-BANDWIDTH=900,CODECS="avc1.77.30"
+http://existing.base/uri/link_2.m3u8
+`,
+		},
```
krystalmejia24, line 1209: Unclear from the PR since only have the diff but do we have a test case where `AVERAGE-BANDWIDTH` is not set and ensure it falls back to bandwidth?

krystalmejia24 (approved):
Small suggestion on an extra test case. otherwise LGTM!

filters/hls_test.go
zsiec: Good catch, added a basic test to cover.

PR #162, opened 2021-08-15
propeller: choose origin explicitly via filters

### LINKED ISSUE #154: propeller: allow selecting specific stream origin urls via an explicit configuration
As a propeller customer, I can specify specific stream origin URLs to be used for clip previews so that the clipping tool does not run into expired segments due to the DAI's rolling manifest window

Acceptance Criteria:

- Bakery supports specified origin URLs via customer configuration
  - e.g Propeller Playback URL vs. GAM DAI URL
- Changes needed in Video Robot are communicated to the P+ Vid Ops team and the UI Squad

Slack thread: [https://cbs.slack.com/archives/C011XPF6E07/p1626291625036700?thread\_ts=1625231494.022800&cid=C011XPF6E07](https://cbs.slack.com/archives/C011XPF6E07/p1626291625036700?thread_ts=1625231494.022800&cid=C011XPF6E07)

Aha! Link: https://viacomcbs-digital.aha.io/features/BAKE-41

zsiec:
### Explicit Origins

To have Bakery use a specific origin url for the output, add another path item specifying the origin to be used:

    http://bakery.dev.cbsi.video/propeller/<org-id>/<channel-id>/<output-id>/<origin>.m3u8

Supported values for `origin` include

|**origin**|**description**|
|----------|---------------|
| dai      | ad-supported stream from Google DAI|
| captions | auto-captioned stream from Dictate |
| cdn      | raw cdn stream url |

#### Origin Fallback

If you're not sure what origins are available for your channel, or to gracefully handle channel configuration
changes that may remove or add origins while the channel is running, you can optionally specify multiple origins
in this last path element. The `-` delimiter separates multiple origins. If the path defines more than one origin,
Bakery will loop through the origins in order and choose the first valid origin found on the channel.

For example, in order to specify a Bakery url that prefers an auto-captioned stream, but falls back to the CDN origin
if auto-captions is disabled on the channel, you'd construct the url:

        http://bakery.dev.cbsi.video/propeller/<org-id>/<channel-id>/<output-id>/captions-cdn.m3u8

fixes #154


origin/propeller.go
```diff
@@ -181,6 +194,10 @@ func (g *outputURLGetter) GetURL(ctx context.Context, client propellerClient) (s
 }
 
 func (g *outputURLGetter) getURL(channel *propeller.Channel, output *propeller.ChannelOutput) (string, error) {
+	if origins := g.origins; len(origins) > 0 {
+		return urlForOrigins(output, origins)
+	}
+
 	if output.AdsURL != "" && channel.Status == "running" {
 		if g.orgID == "uefahda8" {
```
krystalmejia24, line 202: I wonder if we can remove this "hack" here which was the workaround for routing the uefa streams to the originCDN. Then we can enforce the changes on clipping tool for all propeller channels?

zsiec: Good call, we should absolutely circle back to do this once we know how this may be used today and if we can break this implicit contract.

docs/propeller/index.md
```diff
@@ -47,6 +47,31 @@ To request a Propeller output via Bakery:
...
+
+If you're not sure what origins are available for your channel, or to gracefully handle channel configuration
+changes that may remove or add origins while the channel is running, you can optionally specify multiple origins
+in this last path element. The `-` delimiter separates multiple origins. If the path defines more than one origin,
+Bakery will loop through the origins in order and choose the first valid origin found on the channel.
+
+For example, in order to specify a Bakery url that prefers an auto-captioned stream, but falls back to the CDN origin
+if auto-captions is disabled on the channel, you'd construct the url:
+
+        http://bakery.dev.cbsi.video/propeller/<org-id>/<channel-id>/<output-id>/captions-cdn.m3u8
```
krystalmejia24, line 75: Love the new api changes and the way the Propeller paths were extended. When I first thought about this problem myself I was thinking through introducing a filter but didn't like that approach as it seemed like a feature specific to Propeller. With the introduction of fallback and specifying multiple origin channels, longer-term we may be able to use it to route to new origins if one is down.

origin/propeller.go
```diff
@@ -16,11 +17,17 @@ const (
...
+
+	originCDN      = "cdn"
+	originCaptions = "captions"
+	originDAI      = "dai"
 )
 
 // propellerPaths defines the multiple path formats allowed for propeller entities in Bakery
 var propellerPaths = []*regexp.Regexp{
 	regexp.MustCompile(`/propeller/(?P<` + orgIDKey + `>.+)/clip/(?P<` + clipIDKey + `>.+).m3u8`),
+	regexp.MustCompile(`/propeller/(?P<` + orgIDKey + `>.+)/(?P<` + channelIDKey + `>.+)/(?P<` + outputIDKey + `>.+)/(?P<` + originCfg + `>.+).(m3u8|mpd)`),
```
krystalmejia24, line 30: One question I had is if we can use this to support a single output channel as well. If a channel has a single output, I believe the API would look something like this? 

`http://bakery.dev.cbsi.video/propeller/<org-id>/<channel-id>/captions-cdn.m3u8`

krystalmejia24 (changes requested):
Small changes. Also had a question about extending this to support single output channels. I don't know enough about the UEFA channel configurations but if simply extending the multi outputs satisfies the requirement there, then I would be ok with merging this work and adding a ticket to the backlog to extend the explicit origins to single output.

origin/propeller.go
zsiec: Good call, I'll add a ticket to track this. We'll need to be careful here as that pattern above matches the outputID url case and there's no other way to differentiate those path patterns between an output url and single output origin url.

zsiec: https://github.com/cbsinteractive/bakery/issues/164

krystalmejia24 (approved):

origin/propeller.go
athithamarnathcbs: Yes please! good call on that.

Issue #158, opened 2021-07-30
DASH: passthrough content protection attributes being stripped from manifest


krystalmejia24:
We are seeing some attributes (`cenc`, `mspr`) being stripped from the DASH `MPD` element. This is in part because DASH spec expects these attributes to live on the `ContentProtection` element inside the `AdaptationSets`, but Shaka Packager is publishing these attributes on the root `MPD` element,  so these attributes are not being read by `go-dash` library.

In the short term, we'd like to test being able to pass through these attributes on the `MPD` element. Long term solution will be explored in how to read the attribute and properly write them on the `ContentProtection` element.

Note: this was only tested with PlayReady so we may want to look at the outputs for a Widevine manifest from shaka packager as there may be widevine specific tags being stripped as well.

PR #163, opened 2021-08-15
deps: bump head on forked go-dash repo to fix extra attr mapping for drm

zsiec:
Staged changes in our fork: https://github.com/cbsinteractive/go-dash/pull/12

fixes: #158


krystalmejia24 (commented):
Left a comment on the parser changes. Just curious if we need to do anything for tags when dealing with widevine? The `cenc` attribute should be fine but wasn't sure if there were other tags we needed to passthrough. 

Other than that, this looks good and can be merged. We could always do another PR for widevine changes if we need to. Unclear how that works with Shaka myself.

Issue #166, opened 2021-08-31
DASH: update adaptation sets when representations are filtered out


krystalmejia24:
Adaptation Sets have `maxHeight` and `maxWidth` attributes that are derived from the list of representations. When we remove certain media representations from a given adaptation set, we aren't updating the adaptation set to reflect the newly filtered list of representations in the dash manifest.

**REPRODUCE**
Here is a sample manifest from P+
https://vod-gcs-cedexis.cbsaavideo.com/intl_vms/2021/06/24/1913517635745/713198_cenc_dash/stream.mpd

When bandwidth filters are used to limit the max bitrate to 3MB:
https://bakery-pplus-dev.cbs.com/b(0,3000000)/intl_vms/2021/06/24/1913517635745/713198_cenc_dash/stream.mpd

We can see that the largest Representation is now:
```
<Representation id="7" bandwidth="2739766" codecs="avc1.640028" mimeType="video/mp4" sar="1:1" width="1024" height="576">
```

but it is not being reflected in the Adaptation Set:
```
<AdaptationSet id="5" contentType="video" maxWidth="1920" maxHeight="1080" frameRate="24000/1001" segmentAlignment="true" par="16:9">
```

Aha! Link: https://viacomcbs-digital.aha.io/features/BAKE-45

aimeevitacbsi:
Comment added by Aimee Vitaglian in Aha! - [View](https://viacomcbs-digital.aha.io/comments/7008642051011918072)

pushing to sprint 6 due to capacity

PR #172, opened 2021-09-21
calculate max height and width of a given adaptation set after filter…

krystalmejia24:
…ing representations based on bitrate. refs #166


terranvigil (approved):
LGTM. The spec has a few other "shall be present" attributes. Should we add these at the same time? PAR, SAR, FrameRate and others.

vish91:
> LGTM. The spec has a few other "shall be present" attributes. Should we add these at the same time? PAR, SAR, FrameRate and others.

@terranvigil  those other attributes are still there.. as long as the mpd parser can recognize it we wouldn't remove it if the original manifest had those.

PR #176, opened 2022-03-21
Code clean-up, linters, Improve build

terranvigil:


config/config_test.go
```diff
@@ -132,6 +135,7 @@ func TestConfig_LoadConfig(t *testing.T) {
 	}
 
 	for _, tc := range tests {
+		tc := tc
```
joesgit, line 138: just checking if this line is necessary

origin/propeller.go
```diff
@@ -76,12 +87,12 @@ func configurePropeller(ctx context.Context, c config.Config, path string) (Orig
...
 		getter = &channelURLGetter{orgID: orgID, channelID: channelID}
 	}
-	return NewPropeller(ctx, c, getter)
+	return NewPropeller(ctx, conf, getter)
 }
 
 // NewPropeller returns a Propeller origin struct
-func NewPropeller(ctx context.Context, c config.Config, getter urlGetter) (*Propeller, error) {
-	propellerURL, err := getter.GetURL(ctx, &c.Propeller.Client)
+func NewPropeller(ctx context.Context, conf *config.Config, getter urlGetter) (*Propeller, error) {
```
JackDunham, line 94: Does this have implications for thread-safety?

origin/propeller_test.go
```diff
@@ -325,11 +338,13 @@ func TestPropeller_outputURLGetter(t *testing.T) {
 	}
 
 	for _, tc := range tests {
+		tc := tc
```
JackDunham, line 341: Copypasta leftover?

origin/propeller_test.go
```diff
@@ -379,10 +395,12 @@ func TestPropeller_clipURLGetter(t *testing.T) {
...
```
JackDunham, line 398: copypasta, or similar?

JackDunham (commented):
Couple of comments (inline)

go.mod
```diff
@@ -4,19 +4,27 @@ replace github.com/zencoder/go-dash/v3 => github.com/cbsinteractive/go-dash/v3 v
 
 replace github.com/grafov/m3u8 => github.com/cbsinteractive/m3u8 v0.11.2-0.20210702182805-556854f1e40f
 
-go 1.16
+go 1.17
```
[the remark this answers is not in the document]
terranvigil: This was changed by go mod. I'm running 1.17 locally so probably updated when I did a `go mod verify`. But we could certainly change to 1.18.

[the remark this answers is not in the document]
terranvigil: I didn't do any testing with 1.18 but since you're going to be working with it much more, feel free to change when you start your first branch. I'm moving jigsaw to 1.18 when I make my next set of changes since I'll have to do a lot of testing anyway.

origin/propeller.go
terranvigil: I don't think so. Copying response from slack for future reference (or to make sure i get all of the blame for a new bug being introduced :)

```So the config is loaded in cmd/http/main.go:main(). Then it's sent as a pointer to handlers.LoadHandler(). It should only be constructed once per app instance. But it's then used by each request, so yes, the same propellor client instance could be shared by multiple requests from different threads. config/propellor.go is initialized with the same host url no matter how it's called. The class seems threadsafe - even if it was initialized twice the value would be the same. One thing that I would change is that it's manually creating the propellor Client struct instead of using the constructor: https://github.com/cbsinteractive/bakery/blob/master/config/propeller.go#L44 vs https://github.com/cbsinteractive/propeller-go/blob/d02e4f54c5bc20138c257abc787153756f4428a5/client/client.go#L29
And http.Client is threadsafe as well. I'm not seeing any state that could affect thread safety.

[4:36](https://cbs.slack.com/archives/D037Z4XV291/p1648154195000809)
I do think https://github.com/cbsinteractive/bakery/blob/master/config/propeller.go should use pointer receivers for most of its functions since they all reference the Client struct (and so are making copies on each call).
[4:38](https://cbs.slack.com/archives/D037Z4XV291/p1648154287786629)
If we ever change the config to allow multiple propellor hosts, at a minimum, we'd need to change Bakery to use a new client for each propeller host. Hopefully there is a more elegant solution though.```

File: origin/propeller_test.go (hunk 3 above)
Comment (terranvigil) [reply origin/propeller_test.go:341]: Joe had the same question. The tests are run in parallel so this copies the test case to a local variable. Otherwise the parent test will continue running and the test closures will share the same test case data.

File: origin/propeller_test.go (hunk 4 above)
Comment (terranvigil) [reply origin/propeller_test.go:398]: Ditto

File: origin/propeller_test.go (hunk 3 above)
Comment (terranvigil) [reply origin/propeller_test.go:341]: You can see the behavior by running the tests ... all test cases will end up sharing the value of the last declared test case. It's a pretty common mistake to make - common enough that it's documented in a "common go mistakes" document somewhere.

File: filters/dash.go [hunk 6]
```diff
@@ -186,7 +205,7 @@ func (d *DASHFilter) filterAdaptationSetLanguage(filters *parsers.MediaFilters,
 				langs = filters.Audios.Language
 			case captionContentType:
 				langs = filters.Captions.Language
-			default:
+			case videoContentType:
```
vish91, line 208: what's the benefit here ? 
should it be default that would cover both `video` and `image` contentTypes ? both of those anyway don't have a lang descriptor

filters/dash.go
```diff
@@ -287,13 +305,15 @@ func (d *DASHFilter) filterBandwidth(filters *parsers.MediaFilters, manifest *mp
...
 
-			//evaluate bitrate filter for codec type
+			// evaluate bitrate filter for codec type
 			var bitrate *parsers.Bitrate
 			switch ContentType(*as.ContentType) {
 			case videoContentType:
 				bitrate = filters.Videos.Bitrate
 			case audioContentType:
 				bitrate = filters.Audios.Bitrate
+			case captionContentType:
```
vish91, line 315: won't this cause the filter to also remove the dash caption adaptation because in those bandwidth=1000 always ? 
same for image adaptation set ?

parsers/urlparser.go
```diff
@@ -118,117 +133,114 @@ func URLParse(urlpath string) (string, *MediaFilters, error) {
...
 			for _, contentType := range filters {
 				if _, valid := contentSupported[contentType]; !valid {
-					err := fmt.Errorf("Content Type %v is not supported", contentType)
+					err := fmt.Errorf("content Type %v is not supported", contentType)
 					return keyError("Content Type", err)
 				}
-				mf.ContentTypes = append(mf.ContentTypes, contentType)
+				filt.ContentTypes = append(filt.ContentTypes, contentType)
 			}
 		case "l":
```
vish91, line 175: change to constants like others ? and same for other cases in switch statement

filters/dash.go
terranvigil: The benefit is that it's making everything explicit. Right now, video was the only case that was handled by `default:`.  Anything that isn't video, audio or captions wouldn't have a content type and would be handled in the test for nil ContentType. Image doesn't appear to be a declared content type. It could be added to filters/filters.go and related code and then be handled by the switch. Right now I'm guessing it would have a nil content type and be handled just before the switch statement.

filters/dash.go
terranvigil: I was trying to be explicit as in the other case, but we don't support filtering for captions so this code shouldn't be reached in any case. I'll comment out so we review in the future.

parsers/urlparser.go
terranvigil: Oops, I'd created the constants but missed changing them in the switch. Thanks!

vish91: 👍🏽  @terranvigil  there is also a build-vr.yml which we use for the P+ build for bakery. Can you please also update that to match with your build file.

terranvigil: Yes, that will happen as a separate piece of work. Adding @JackDunham

Issue #182, opened 2022-03-24
Bakery: V2: Add DVS (Descriptive Video Service) and ATMOS audio filter values


robomac:
Note: For a more readable description, reference the Bakery v2 Specification

Several new filter values are added:
### Descriptive Video Service (DVS)
DVS is an audio track describing the action, generally timed to fit into pauses in the the dialog. This is also called "audio description", and, in HLS, is tagged with CHARACTERISTICS="public.accessibility.describes-video"
Filtering of this type in audio tracks is added.
### ATMOS Support
ATMOS is an extension of EAC-3 audio, “ec-3”. This will be called out as a new audio filter value, “atm”. Care must be taken to ensure that:
 The ec-3 filter does not remove ATMOS streams despite ATMOS being a subset of “ec- 3”.
 The ATMOS (“`atm”) filter does not impact non-ATMOS EAC-3 (“ec-3”) streams.
In DASH, it is represented by additional “supplementalProperty” tags flagging it as JOC, as in:
<SupplementalProperty schemeIdUri="tag:dolby.com,2018:dash:EC3_ExtensionType:2018" value="JOC"/> <SupplementalProperty schemeIdUri="tag:dolby.com,2018:dash:EC3_ExtensionComplexityIndex:2018" value="16"/>
As in the example for the second audio stream below:
<Representation id="0" bandwidth="194216" codecs="ec-3" mimeType="audio/mp4"
audioSamplingRate="48000">
<AudioChannelConfiguration schemeIdUri="tag:dolby.com,2014:dash:audio_channel_configuration:2011" value="F801"/> <BaseURL>Halo_101_English_Audio_51_Rev1_en-US_1243566_1_imf_eac3_192.mp4</ BaseURL> <SegmentBase indexRange="1472-8583" timescale="48000">
<Initialization range="0-1471"/>
</SegmentBase>
</Representation>
<Representation id="1" bandwidth="770216" codecs="ec-3" mimeType="audio/mp4" audioSamplingRate="48000">
<AudioChannelConfiguration schemeIdUri="tag:dolby.com,2014:dash:audio_channel_configuration:2011" value="F801"/> <SupplementalProperty schemeIdUri="tag:dolby.com,2018:dash:EC3_ExtensionType:2018" value="JOC"/>
[...]

PR #191, opened 2022-03-30
Bakery v2: filter options (-f, -i), atmos filtering, dvs filtering

JackDunham:
**Summary**
According to [Bakery 2.0 Spec
](https://www.notion.so/videotechnologygroup/Bakery-Internal-b13753ccc9074d9598c5c3d387abac4c#1ec9c3598089494392a1b9d0a5383706)

Implement:
- `include` and `include-first` filter-options
- Atmos filtering
- Descriptive Video Service (DVS) filtering

**Linked Issues**
#182
#178


terranvigil:
Looks like some tests are failing. Also, if you merge in my branch or master, those annoying inline comments from codecov will go away.

filters/dash.go
```diff
@@ -154,26 +154,46 @@ func (d *DASHFilter) filterCaptionTypes(filters *parsers.MediaFilters, manifest
...
+	filterContentType(captionContentType, supportedCaptionTypes, manifest, filters.Captions.FilterOption)
 }
 
-func filterContentType(filter ContentType, supportedContentTypes map[string]struct{}, manifest *mpd.MPD) {
+func filterContentType(filter ContentType, supportedContentTypes map[string]struct{}, manifest *mpd.MPD,
+	filterOption parsers.FilterOption) {
 	for _, period := range manifest.Periods {
 		var filteredAdaptationSets []*mpd.AdaptationSet
 		for _, as := range period.AdaptationSets {
+			// TODO(jack): do stuff here, depending on option
```
terranvigil, line 165: Is the TODO work tracked by a separate ticket? Generally, approving a PR means it's ready to be merged to main and go to dev. But if this is more of a review to get feedback, then we prefix the PR name with "WIP:" and ignore the TODOs and just make comments but don't approve the PR.

filters/dash.go
```diff
@@ -154,26 +154,46 @@ func (d *DASHFilter) filterCaptionTypes(filters *parsers.MediaFilters, manifest
...
 				for _, rep := range as.Representations {
 					if rep.Codecs == nil {
 						filteredReps = append(filteredReps, rep)
 						continue
 					}
+					if filterOption == parsers.FilterOptionIncludeFirst {
+						if matchCodec(*rep.Codecs, filter, supportedContentTypes) {
+							// Include of first of any matching/specified codes.
+							// After the first matching codec has been found, ignore any
+							// subsequent matching codes
```
terranvigil, line 178: codecs? SP

terranvigil:
Code looks good and great tests. CI is failing due to linters.

JackDunham:
Testing PR notification emails.

filters/dash.go
JackDunham: Crufty comment, actually.

JackDunham: Fixed

filters/dash.go
JackDunham: Fixed

filters/dash.go
```diff
@@ -154,26 +158,45 @@ func (d *DASHFilter) filterCaptionTypes(filters *parsers.MediaFilters, manifest
...
+					if filterOption == parsers.FilterOptionIncludeFirst {
+						if matchCodec(*rep.Codecs, filter, supportedContentTypes) {
+							// Include of first of any matching/specified codecs.
+							// After the first matching codec has been found, ignore any
+							// subsequent matching codes
+							if !foundFirst {
+								filteredReps = append(filteredReps, rep)
+								foundFirst = true
+							}
+							continue
```
terranvigil, line 187: Shouldn't this break instead of continue?

filters/dash.go
```diff
@@ -200,18 +223,73 @@ func (d *DASHFilter) filterAdaptationSetLanguage(filters *parsers.MediaFilters,
 			}
 
 			var langs []string
-			switch ContentType(*as.ContentType) {
```
terranvigil, line 203: Just curious, why the change from switch to if/else? The former seems more concise.

filters/dash.go
```diff
@@ -200,18 +223,73 @@ func (d *DASHFilter) filterAdaptationSetLanguage(filters *parsers.MediaFilters,
...
 				langs = filters.Captions.Language
-			case videoContentType:
+				filterOption = filters.Captions.FilterOption
+			} else {
 				filteredAdaptationSets = append(filteredAdaptationSets, as)
 				continue
 			}
 
-			if !(matchLang(*as.Lang, langs)) {
+			if filterOption == parsers.FilterOptionIncludeFirst { // nolint:nestif
```
terranvigil, line 239: If this linter isn't providing help, maybe add to the ignore list? We can always move it back and see if it's worth restructuring nested if statements.

filters/dash.go
JackDunham: One guess: lint was complaining.

filters/dash.go
JackDunham: As w/ everything lint -- I don't mind keeping it on, and occasionally disabling it, on an as-needed basis. If we're disabling it everywhere, then -- yes -- should add ignore-list.

filters/dash.go
JackDunham: Nope -- we still need to process the all the representations in the AdaptationSet -- we just don't need to bother w/ the subsequent logic, in the loop.

filters/dash.go
```diff
@@ -154,26 +158,71 @@ func (d *DASHFilter) filterCaptionTypes(filters *parsers.MediaFilters, manifest
...
 					if rep.Codecs == nil {
 						filteredReps = append(filteredReps, rep)
 						continue
 					}
 
-					if matchCodec(*rep.Codecs, filter, supportedContentTypes) {
-						continue
+					// Replace original codecs with "special" codecs (i.e. ec-3 -> atm, if rep is actually atmos)
+					repCodecs := handleSpecialCodecs(filter, rep)
+
```
terranvigil, line 203: Could also use a switch statement to avoid the comparisons.

filters/dash.go
terranvigil: Ah, is that because of the "handleSpecialCodecs" piece?

filters/dash.go
terranvigil: Hmm, it shouldn't. I know it will complain if you only have a single "case" but that's not true here.

filters/dash_test.go
```diff
@@ -806,6 +807,319 @@ func TestDASHFilter_FilterRole_OverwriteValue(t *testing.T) {
 	}
 }
 
+func TestDASHFilter_FilterContent_Include(t *testing.T) {
+	t.Parallel()
+
+	manifestWithAccessibilityElement := readTestFixtures("manifest_with_accessibility_element.mpd")
```
terranvigil, line 813: thank you for not putting the manifests inline ... so much easier to read when they area test fixture!

filters/dash_test.go
```diff
@@ -806,6 +807,319 @@ func TestDASHFilter_FilterRole_OverwriteValue(t *testing.T) {
...
+		filters               *parsers.MediaFilters
+		manifestContent       string
+		expectManifestContent string
+		expectErr             bool
+	}{
+		{
+			name: "when proper value is set and manifest has accessibility element, role value is overwritten.",
+			filters: &parsers.MediaFilters{
+				Videos: parsers.NestedFilters{
+					Codecs:       []string{"avc", "dvh"},
```
terranvigil, line 905: Could all of these values be constants in filters/common.go?

filters/hls.go
```diff
@@ -183,14 +195,33 @@ func (h *HLSFilter) filterPipeline(ctx context.Context, uri string) (pipelineTyp
 	return backupPipeline, nil
 }
 
+func (h *HLSFilter) filterRepresentations(filters *parsers.MediaFilters, vari *m3u8.Variant) bool {
+	filtered := false
+	for idx, alternative := range vari.VariantParams.Alternatives {
+		for _, chr := range filters.Characteristics {
+			switch chr {
+			case "dvs":
```
terranvigil, line 203: Now this is where I'd expect a linter to complain. Could be changed to:
```
if chr == "dvs" && alternative.Characteristics == dvsCharacteristics {
	vari.VariantParams.Alternatives = append(vari.VariantParams.Alternatives[:idx], vari.VariantParams.Alternatives[idx+1:]...)
	filtered = true
	break
}
```

filters/hls.go
```diff
@@ -199,46 +230,34 @@ func (h *HLSFilter) filterVariant(filters *parsers.MediaFilters, vari *m3u8.Vari
...
-		supportedVideoTypes := map[string]struct{}{}
-		for _, vt := range filters.Videos.Codecs {
-			supportedVideoTypes[vt] = struct{}{}
-		}
-		res, err := filterVariantCodecs(videoContentType, variantCodecs, supportedVideoTypes, matchFunctions)
-		if res {
-			return true, err
-		}
+	// Filter video
+	res, filterOptionsState, err := filterByContentType(videoContentType, filters.Videos.Codecs, variantCodecs,
```
terranvigil, line 238: Don't have a strong opinion on this one but wasn't sure if you knew the other form for this:
```
if res, filterOptionsState, err := filterByContentType(videoContentType, ... filterOptionsState); res {
	return res, filterOptionsState, err
}
```

filters/hls.go
```diff
@@ -199,46 +230,34 @@ func (h *HLSFilter) filterVariant(filters *parsers.MediaFilters, vari *m3u8.Vari
...
-		supportedAudioTypes := map[string]struct{}{}
-		for _, at := range filters.Audios.Codecs {
-			supportedAudioTypes[at] = struct{}{}
-		}
-		res, err := filterVariantCodecs(audioContentType, variantCodecs, supportedAudioTypes, matchFunctions)
-		if res {
-			return true, err
-		}
+	// Filter audio
+	res, filterOptionsState, err = filterByContentType(audioContentType, filters.Audios.Codecs, variantCodecs,
```
terranvigil, line 245: I haven't read far enough to understand why this would make sense but without the context or comment, it's confusing to see `filters.Videos.FilterOption` used in filtering audio.

filters/hls.go
```diff
@@ -199,46 +230,34 @@ func (h *HLSFilter) filterVariant(filters *parsers.MediaFilters, vari *m3u8.Vari
...
-		supportedCaptions := map[string]struct{}{}
-		for _, ct := range filters.Captions.Codecs {
-			supportedCaptions[ct] = struct{}{}
-		}
-		res, err := filterVariantCodecs(captionContentType, variantCodecs, supportedCaptions, matchFunctions)
-		if res {
-			return true, err
-		}
+	// Filter captions
+	res, filterOptionsState, err = filterByContentType(captionContentType, filters.Captions.Codecs, variantCodecs,
```
terranvigil, line 252: And here

filters/hls.go
```diff
@@ -248,7 +267,42 @@ func (h *HLSFilter) filterVariant(filters *parsers.MediaFilters, vari *m3u8.Vari
...
+	filterOptionsState FilterOptionsState,
+) (bool, FilterOptionsState, error) {
+	if filterCodecs != nil { //nolint:nestif
+		supportedTypes := map[string]struct{}{}
+		for _, vt := range filterCodecs {
+			supportedTypes[vt] = struct{}{}
+		}
+		res, filteredCodec, err := filterVariantCodecs(contentType, variantCodecs, supportedTypes, matchFunctions)
+
+		if filterOptions == parsers.FilterOptionInclude {
```
terranvigil, line 287: Could && the compares here and below to avoid the extra nesting.

logging/logging_test.go
```diff
@@ -0,0 +1,9 @@
+package logging
+
+import (
+	"testing"
+)
+
+// No-op test
```
terranvigil, line 7: Is this just a place-holder for future tests? Otherwise no value.

parsers/urlparser.go
```diff
@@ -135,33 +149,61 @@ func URLParse(urlpath string) (string, *MediaFilters, error) {
...
+		} else if partsDict["option"] == "f" {
+			filterOption = FilterOptionIncludeFirst
+		}
+
+		switch partsDict["key"] {
+		case ParamCharacteristics:
+			for _, nf := range nestedFilters {
+				found := false
+				for k := range characteristicsSupported {
+					if k == nf {
```
terranvigil, line 182: Simplify by doing:
```
if _, found := characteristicsSupported[nf] {
    break;
} else {
    return keyError("Characteristics", nil)
}
```

parsers/urlparser.go
```diff
@@ -298,6 +344,12 @@ func (nf *NestedFilters) parseKeys(key string, values []string) error {
 		for _, val := range values {
 			switch val {
 			case "hdr10":
+				// NOTE: this is technically wrong, in the sense that hev1.2 and hvc1.2 content are only HDR if:
+				// For HLS: VIDEO_RANGE="PG" or "HLG"
+				// or
+				// For DASH: profle='http://dashif.org/guidelines/dash-if-uhd#hevc-hdr-pq10' AND
+				//           codecs = "hvc1.2.4.L153.B0" or "hev1.2.4.L153.B0"
+				// For our purposes, this is OK (currently) because (per Jignesh) all HEVC content is assumed to be HDR
```
terranvigil, line 352: This just doesn't feel right. Jignesh isn't our only customer and we will continue to add customers.

tests/tests_test.go
```diff
@@ -0,0 +1,9 @@
+package tests
```
terranvigil, line 1: ???

terranvigil (changes requested):
Overall, looks great. Had a couple comments, questions and suggestions.

filters/dash.go
JackDunham: That and...part of how include vs exclude now works.

filters/dash.go
terranvigil: Ah, that's lame. That's the whole point of default. I do get why you might want to not rely on a default and force new possible values to force revisiting the logic, but there is probably a better way.

filters/dash_test.go
JackDunham: 👍

filters/hls.go
JackDunham: Oh my, no! That's gotta be an error...Why does it pass all the tests...? Good catch

JackDunham: Where is that happening?

JackDunham: Ah, yeah -- that's all fine. It's because:

JackDunham: Included a comment, re: use/purpose of FilterOptionsState

filters/hls.go
JackDunham: See comment attached to `filterVariant` method, re: use/purpose of FilterOptionsState

filters/hls.go
JackDunham: This is true...but I don't know that it makes the code any more readable. IMHO, judicious use of nested IFs can be _more_ readable.

parsers/urlparser.go
JackDunham: Here's my thinking (here and above, where you suggested removing the range over characteristics): I can easily imagine a short-term future where there's more than one characteristic we need to handle.

parsers/urlparser.go
JackDunham: Oh, for sure it's not "right". But it's def. what @jigneshdhruv indicated that _want_. (They _know_ it's not "right". That also promise any hvc1.2/hev1.2 we see _will_ be 10-bit color).

tests/tests_test.go
JackDunham: More monkey-tricks for getting this linter to stop complaining about no tests for a module...even when that module is a test-fixture directory.

JackDunham: It's definitely something than can be sorted out in the `Makefile`...

filters/dash.go
JackDunham: Yeah...linters in general seem to err on overly-aggressive side...¯\_(ツ)_/¯ Everything taken w/ a grain of salt...

filters/dash.go
terranvigil:
```
switch (filterOption) {
}
```
And then //nolint:whatever that is complaining about not listing out all options

parsers/urlparser.go
terranvigil: My bias is based on our tendency to treat each new project as an emergency and never get back to filling in features, leaving lots of TODOs and FIXMEs in the code. But I hear you.

parsers/urlparser.go
terranvigil: hvc/hev1.2.4 will always bee Main10 and therefore 10-bit. Still makes me really uncomfortable. Maybe add a note to the docs about this one so we can at least says  "We warned you".

filters/dash_test.go
JackDunham: Fixed.

parsers/urlparser.go
```diff
@@ -135,33 +149,61 @@ func URLParse(urlpath string) (string, *MediaFilters, error) {
...
 		}
 
-		filters := strings.Split(subparts[2], ",")
-		nestedFilters := splitAfter(subparts[2], nestedFilterRegexp)
+		filters := strings.Split(partsDict["inner"], ",")
+		nestedFilters := splitAfter(partsDict["inner"], nestedFilterRegexp)
 
-		switch key := subparts[1]; key {
+		filterOption := FilterOptionExclude
+		if partsDict["option"] == "i" {
```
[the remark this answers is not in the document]
JackDunham: Fixed.

parsers/urlparser.go
JackDunham: Found a better way to express this, actually.  Calling it fixed.

filters/hls.go
JackDunham: Yeah --I'm definitely _aware_ of it. And I know it's one of the "Go-ier" of all the Go idioms. I just don't find it makes anything more readable.

filters/dash.go
JackDunham: Changed.

filters/hls.go
JackDunham: Changed.

terranvigil (approved):
LGTM!

